PR #11, opened 2015-09-22
Use 'python2.7' in shebang; make script executable

sherter:
On modern systems 'python' is most likely symlinked to python3, but
'google-java-format-diff.py' uses python2 syntax. Therefore it is better
to explicitly use 'python2'. Also changed file mode of the script (chmod
a+x) to make it executable.

### DIFF
--- a/scripts/google-java-format-diff.py
+++ b/scripts/google-java-format-diff.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 #
 #===- google-java-format-diff.py - google-java-format Diff Reformatter -----===#
 #


sherter:
I signed it!

cushon:
Thanks for the patch, but I don't think it works on macs. Any idea how portable `#!/usr/bin/env python2.7` is?

This script was adapted from [clang-format-diff](https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/clang-format-diff.py), which uses the same shebang. I wonder if it's causing problems there too.

sherter:
I quickly checked a few distributions (Ubuntu 15.04., Fedora 22, Debian 8.1, Arch) and all provide the `python2.7` symlink. Are you saying `python2.7` works on your mac?

cushon:
> Are you saying #!/usr/bin/env python2.7 works on your mac?

Yep, I think all recent versions of OS X ship with a python2.7 symlink, but not python2.

PR #17, opened 2015-12-10
google-java-format-diff.py: improve finding of executable

scottbessler:
use python's ability to find an executable rather than assuming the executable is in /usr/bin


scottbessler:
I signed it!

cushon:
Thanks!

jontg:
:boom:

PR #21, opened 2016-01-14
Make the shaded jar a secondary artifact

### LINKED ISSUE #20: Provide an artifact that shades or omits Guava and other dependencies
As noted in https://github.com/google/dagger/issues/228#issuecomment-171242072 and the comment it replies to, the `google-java-format` jar contains classes from Guava and other dependencies. Those classes end up in Dagger, where they conflict with the version of Guava used by Dagger -- and, because of Maven's lack of support for `-processorpath`, with the version of Guava used by Dagger *users* like Caliper.

I don't know enough about Maven to say what the right thing here is. The easiest thing, I'd guess, is shading, but for all I know, this may be difficult with the Eclipse classes for some reason.

cushon:
Fixes #20

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -236,35 +236,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-shade-plugin</artifactId>
-        <version>2.1</version>
-        <executions>
-          <execution>
-            <phase>package</phase>
-            <goals>
-              <goal>shade</goal>
-            </goals>
-            <configuration>
-              <shadedArtifactAttached>true</shadedArtifactAttached>
-              <shadedClassifierName>deploy</shadedClassifierName>
-              <createDependencyReducedPom>false</createDependencyReducedPom>
-              <!-- http://stackoverflow.com/a/6743609 -->
-              <filters>
-                <filter>
-                  <artifact>*:*</artifact>
-                  <excludes>
-                    <exclude>META-INF/*.SF</exclude>
-                    <exclude>META-INF/*.DSA</exclude>
-                    <exclude>META-INF/*.RSA</exclude>
-                  </excludes>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -247,6 +247,8 @@
               <goal>shade</goal>
             </goals>
             <configuration>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
```
tbroyer: This will also affect the IDEA plugin **and** the parent (it generates a `google-java-format-parent-0.1-SNAPSHOT-deploy.pom` file as an empty zip archive that *will* be deployed to Central).

IMO, that plugin (unless needed for the IDEA plugin too) should be moved to `core/pom.xml` (or possibly to `<pluginManagement>` and then added explicitly –without configuration– into `core/pom.xml`)

Otherwise, yes, the generated JAR for `core` is OK.

pom.xml
```diff
@@ -247,6 +247,8 @@
...
+              <shadedClassifierName>deploy</shadedClassifierName>
```
tbroyer: I'm not a fan of that name. How about `jar-with-dependencies` like the `maven-assembly-plugin`'s [predefined descriptor](https://maven.apache.org/plugins/maven-assembly-plugin/descriptor-refs.html#jar-with-dependencies) would do?
(which is also why I suggested using the `maven-assembly-plugin` here)

Or maybe `all-deps` or `alldeps`? or `standalone`?

pom.xml
cushon: Thanks, I moved it to core.

pom.xml
cushon: I renamed to `all-deps`, 'deploy' jars are a bazel-ism: http://bazel.io/docs/be/java.html#java_binary_implicit_outputs

cushon:
merged as 1d28d382fc316f27ae455254277e9d33f245ee96

PR #26, opened 2016-02-16
Provide programmatic access to error details

### LINKED ISSUE #25: Provide programmatic access to fields in FormatterDiagnostic
I know this project is in an early stage and you are probably working on the public API internally. However, I just stumbled upon this issue and wanted to leave a note.

One can retrieve the [diagnostics](https://github.com/google/google-java-format/blob/master/core/src/main/java/com/google/googlejavaformat/java/FormatterException.java#L37) from a FormatterException, but a [FormatterDiagnostic](https://github.com/google/google-java-format/blob/master/core/src/main/java/com/google/googlejavaformat/FormatterDiagnostic.java) doesn't provide programmatic access to the error information. This forces the programmer to parse the String returned by [toString()](https://github.com/google/google-java-format/blob/master/core/src/main/java/com/google/googlejavaformat/FormatterDiagnostic.java#L39) (Effective Java Item 10).

sherter:
The toString() method in FormatterDiagnostic.java returns detailed
information about where an error occurred. This commit adds getter
methods to make this information available programmatically.

Closes #25

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/FormatterDiagnostic.java
+++ b/core/src/main/java/com/google/googlejavaformat/FormatterDiagnostic.java
@@ -35,6 +35,26 @@ public class FormatterDiagnostic {
     this.message = message;
   }
 
+  /** Returns the name of the file that was processed. */
+  public String filename() {
+    return filename;
+  }
+
+  /** Returns the line number on which the error occurred. */
+  public int line() {
+    return lineNumber;
+  }
+
+  /** Returns the 0-indexed column number on which the error occurred. */
+  public int column() {
+    return column;
+  }
+
+  /** Returns a description of the problem that prevented formatting from succeeding. */
+  public String message() {
+    return message;
+  }
+
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();


core/src/main/java/com/google/googlejavaformat/FormatterDiagnostic.java
```diff
@@ -35,6 +35,26 @@ public FormatterDiagnostic(String filename, int lineNumber, int column, String m
...
+  public String filename() {
+    return filename;
+  }
+
+  /** Returns the line number on which the error occurred. */
+  public int line() {
+    return lineNumber;
+  }
+
+  /** Returns the column number on which the error occurred. */
```
cushon: nit: maybe "Returns the 0-indexed column number ..." since it's not obvious which to expect.

cushon:
Thanks, LGTM

core/src/main/java/com/google/googlejavaformat/FormatterDiagnostic.java
sherter: updated and squashed in 9159a98f8040ca40dcaddb577ffcf65af06af56f

PR #32, opened 2016-03-25
Use Files.move instead of File.renameTo

sormuras:
`--replace` does not work on Windows, yields `cannot rename temp file`. After switching to `java.nio.Files.move()`, it works here, too.

See http://stackoverflow.com/questions/1000183/reliable-file-renameto-alternative-on-windows for details.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/FormatFileCallable.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/FormatFileCallable.java
@@ -34,6 +34,8 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


sormuras:
I signed it!

core/src/main/java/com/google/googlejavaformat/java/FormatFileCallable.java
```diff
@@ -232,7 +234,12 @@ private boolean writeOutput(Write write) {
...
       }
-      if (!new File(tempFileName).renameTo(new File(fileToFormat.fileName()))) {
+      try {
+        Files.move(
+            new File(tempFileName).toPath(),
+            new File(fileToFormat.fileName()).toPath(),
+            StandardCopyOption.REPLACE_EXISTING);
+      } catch(IOException e) {
         synchronized (outputLock) {
           errWriter.append(tempFileName).append(": cannot rename temp file").append('\n').flush();
```
cushon: Maybe log the exception message, to be consistent with the error handling in this method?

cushon:
Thanks! Looks good, aside from my other comment.

Did you run into any other issues? We want to support windows, but it hasn't gotten much testing there yet. I think we're baking in assumptions about line terminators, and maybe other things?

sormuras:
Ja, the line ending/whitespace issue showed up once. Eclipse/egit reported a lot of "ignorable whitespace" differences when running the formatter via "Run...", "Main" and "--replace ${selected_resource_loc}" on a singe file. Need to check the result, again.

I automated the formatting task in an Ant target, which did not produce those line ending confusions in Eclipse:
```xml
  <target name="format" description="apply google-java-format">
    <apply executable="${java.home}/bin/java">
      <arg line="-jar lib/build/google-java-format-0.1-SNAPSHOT-all-deps.jar" />
      <arg line="--replace"/>
      <srcfile />
      <fileset dir="src" includes="**/*.java" />
    </apply>
  </target>
```

PR #34, opened 2016-03-28
Preserve "fileToFormat" in replace mode.

sormuras:
If the formatter is in replace mode and changed nothing (ignoring line terminators), don't replace the original file by the temporary one.

I'll add a test case, if that behaviour is wanted. Also, add an option to force move the temp file and bypass the comparison.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/FormatFileCallable.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/FormatFileCallable.java
@@ -25,7 +25,6 @@ import com.google.googlejavaformat.FormatterDiagnostic;
 import com.google.googlejavaformat.java.JavaFormatterOptions.SortImports;
 
 import java.io.BufferedOutputStream;
-import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


cushon:
I'm not sure this is the right fix. We want to do something sensible with line terminators even if formatting produces changes. One suggestion was to always use the platform line terminator. We could also try to detect the existing line terminator and preserve it. Thoughts?

Also, it'd be nice to avoid reading in the files twice.

sormuras:
This PR, at the cost of reading the files twice, keeps the original file untouched. Last modification date and other properties are not altered - as the original file isn't touched at all. Whether the temporary file uses the same or a different line terminator doesn't matter in this case.

I use **google-java-format** as an external tool within Eclipse (executed via an Ant task) at the moment. With the current "always replace original" behaviour, all opened files needs to be updated (hit F5) even if nothing changed.

Do you see another way to keep the `fileToFormat` untouched?

cushon:
Only writing files if the contents change sounds good, but it should be easy to do that without reading the file twice. I would prefer to have FormatFileCallable just save the input from the first read to do the comparison at the end.

sormuras:
Is it correct assuming if in `JavaOutput#writeMerged` the `replacements` list is empty, there's no change? Or is this too deep and should not be used?

On the other hand, `JavaInput#text` could be passed to the `FormatFileCallable#writeOutput` method, converted to a `List<String>` there and used for the comparison.

PR #35, opened 2016-03-31
Preserve original file in replace mode if nothing changed.

sormuras:
Follow up from #34 using already present lines.


sormuras:
Local tests running against all `*.output` test files are almost green... needs more work, though.

sormuras:
Superseded by #41

PR #40, opened 2016-04-13
Fixes tests under Windows.

sormuras:
Which one do you prefer?


cushon:
I think `contains` is fine, thanks.

sormuras:
`contains` prevailed, squashed, you're welcome.

PR #41, opened 2016-04-13
Preserve original file in replace mode if nothing changed.

sormuras:


core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -207,6 +207,7 @@ private int formatFiles(ArgInfo argInfo, JavaFormatterOptions options) {
     int numThreads = Math.min(MAX_THREADS, argInfo.parameters.fileNamesFlag.size());
     ExecutorService executorService = Executors.newFixedThreadPool(numThreads);
 
+    Map<Path, String> inputs = new LinkedHashMap<>();
```
cushon: Instead of creating a second map, what do you think about changing `FormatFileCallable` to return a result that includes information about whether the file changed as well as the formatted output?

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -69,4 +74,28 @@ public void cantImportSortRangeYet() throws UsageException {
     assertThat(err.toString()).contains(
         "--sort-imports can currently only apply to the whole file");
   }
+
+  @Test
+  public void preserveOriginalFile() throws Exception {
+    Path from = Paths.get("src/test/resources/com/google/googlejavaformat/java/testdata/");
+    Path temp = testFolder.newFolder("preserve").toPath();
+    try (DirectoryStream<Path> stream = Files.newDirectoryStream(from, "*.output")) {
```
cushon: Testing on a single file would probably be sufficient?

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Sure, didn't want to obscure the logic behind the simple 3-line implementation.

What would the result look like? Return `FormatFileCallable::this` and the caller cares to extract the information needed - or a dedicated `FormatResult` class that only contains the needed information like `String formatted` and `boolean changed`. What else?

cushon: I was imagining something like `FormatResult` as a nested class in `FormatFileCallable`, but this is simple enough that I'm not sure it's worth it. I'll leave it up to you.

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
sormuras: But it's always good to know, that all generated/expected outputs are congruent with themselves, not? Is the duration too long? Or will it be some day, when there're 1000s of `*.output`files?

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: I tend to keep the second map. It's a simple use-case, the caller has all information needed and the callable has no benefit from known if something changed.

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
cushon: I'm not really worried about performance, and testing that the formatter is idempotent is a good idea.

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -69,4 +74,28 @@ public void cantImportSortRangeYet() throws UsageException {
...
+    }
+  }
+
+  private void preserveOriginalFile(Path java) throws Exception {
+    long expected = Files.getLastModifiedTime(java).toMillis();
+    StringWriter out = new StringWriter();
+    StringWriter err = new StringWriter();
+    Main main = new Main(new PrintWriter(out, true), new PrintWriter(err, true), System.in);
+    int errorCode = main.format("-replace", java.toAbsolutePath().toString());
+    Assert.assertEquals("ErrorCode not 0: " + err + ": " + out, 0, errorCode);
```
cushon: Consider using truth assertions here for consistency with the rest of the file:

```java
assertThat(errorCode).named("Error Code").isEqualTo(0);
```

cushon:
This LGTM, I can merge it tomorrow.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: One last thought: what about changing the existing map collecting `FormatResult` instead of `Future<String>`? With `FormatResult` having a pointer to the input and the callable?

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
sormuras: Done.

sormuras:
For the record and consideration: I implemented the single map in `Main` -- and still like #41 more.

https://github.com/google/google-java-format/compare/master...sormuras:preserve_original_one_map?expand=1

What do you think?

cushon:
Merged, thanks.

PR #45, opened 2016-05-04
Fix printing usage info

amc6:
The try-with-resources block is causing stdout and stderr to be closed before the catch clause. As a result usage info is never printed. The easiest fix is to split into two separate try blocks.


cushon:
Thanks!

Issue #42, opened 2016-04-19
SortImports option is ignored


sherter:
No matter which `SortImports` option I set (`ONLY`, `ALSO` or `NO`), the imports are never sorted and it always returns the same result. Am I missing something here? 
```
public static void main(String[] args) throws FormatterException {
  JavaFormatterOptions options = new JavaFormatterOptions(JavadocFormatter.NONE, Style.GOOGLE, SortImports.ONLY);
  Formatter formatter = new Formatter(options);
  String result = formatter.formatSource("package com.google.example;\n" +
                "\n" +
                "import com.google.common.base.Preconditions;\n" +
                "\n" +
                "import org.junit.runner.RunWith;\n" +
                "import org.junit.runners.JUnit4;\n" +
                "\n" +
                "import java.util.List;\n" +
                "\n" +
                "import javax.annotations.Nullable;\n" +
                "\n" +
                "import static org.junit.Assert.fail;\n" +
                "import static com.google.truth.Truth.assertThat;\n" +
                "\n" +
                "@RunWith( JUnit4.class ) public class SomeTest  {}");
  System.out.println(result);
}
```

sherter:
Seems like the `Formatter` can't do imports sorting at all. This is all done in `FormatFileCallable` with a different `SortImports` enum, which is not part of the public API. The principle of **most** astonishment definitely applies here :wink:
(to be honest, the whole API doesn't feel like a `1.0` version, more like a `0.1`.... typo? :smile:)

cushon:
Yep, currently import ordering only works with the cli. It's work in progress.

Adding it to the methods in the api that return strings would be easy, but some details need to be sorted out for the methods that do partial formatting or return replacement ranges.

talios:
Look forward to seeing this in a release, so I can also plug it into my https://github.com/talios/googleformatter-maven-plugin Maven plugin.

sherter:
The same goes for the `JavadocFormatter` option. JavaDoc is _always_ formatted, even if the formatter is set to `JavadocFormatter.NONE`. I'm wondering why you make an API public although it's not providing any functionality.

cushon:
>  JavaDoc is always formatted, even if the formatter is set to JavadocFormatter.NONE.

Do you have a repro?

sherter:
I wouldn't expect the text to change in any way in the example below. Or is this not how it works?

```
public static void main(String[] args) throws FormatterException {
[...]

PR #50, opened 2016-05-24
'jfmt' tool for the fast formating

pvlbzn:
I wrote a tool, Python script, which is able to format .java files in directory and subdirectories or just a single .java file. I use it like `$ jfmt /path/to/project/dir` and I find it very convenient way to format my code as a whole.

Script was tested on Linux and Mac.

I hope this tool can be useful for others. Also, I not quite sure about the name. I was inspired by `fmt` from the Go lang.

### DIFF
--- a/scripts/jfmt.py
+++ b/scripts/jfmt.py
@@ -2,7 +2,7 @@
 
 from __future__ import print_function
 import sys, os
-
+import argparse
 
 # Purpose:
 # Format all .java files in the directory, including child
@@ -38,10 +38,9 @@ def origin_path():
 GJF = "google-java-format-1.0-all-deps.jar"
 # Command which will be executed by the os.system()
 COMMAND = "java -jar " + origin_path() + "/" + GJF
-# Verbose output.
-DEBUG = False
 
-def format_files(p):
+
+def format_files(p, verbose):
     '''
     Format files in `p` directory.
     '''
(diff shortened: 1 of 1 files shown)


kevinb9n:
Thanks for thinking about how to make this project better!

Question: If `google-java-format` could take directories as arguments and
process them recursively, would a separate wrapper be necessary?


On Tue, May 24, 2016 at 5:37 AM, Pavel Bazin <notifications@github.com>
wrote:

> I wrote a tool, Python script, which is able to format .java files in
> directory and subdirectories or just a single .java file. I use it like $
> jfmt /path/to/project/dir and I find it very convenient way to format my
> code as a whole.
>
> Script was tested on Linux and Mac.
>
> I hope this tool can be useful for others. Also, I not quite sure about
> the name. I was inspired by fmt from the Go lang.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/google/google-java-format/pull/50
> Commit Summary
>
>    - Python script 'jfmt' for formatting automatization.
>    - Fixed comment in jfmt.
>    - Added 'Purpose' description to 'jfmt'.
>
> File Changes
>
>    - *A* scripts/jfmt.py
>    <https://github.com/google/google-java-format/pull/50/files#diff-0>
>    (103)
>
> Patch Links:
>
>    - https://github.com/google/google-java-format/pull/50.patch
>    - https://github.com/google/google-java-format/pull/50.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/google/google-java-format/pull/50>
>



-- 
Kevin Bourrillion | Java Librarian | Google, Inc. | kevinb@google.com

sherter:
>Question: If `google-java-format` could take directories as arguments and
process them recursively, would a separate wrapper be necessary?

I had this idea before, but as you suggest to put more functionality into the executable I think it should be considered now: What do you think about separating library code and executable into two artifacts? This would prevent problems like https://github.com/google/google-java-format/issues/42 in the future and tools like dagger could depend only on the library. CLI parsing libraries used in the executable are transitively resolved right now, whether you use google-java-format as standalone tool or only the `Formatter`.

pvlbzn:
If you ask me, I'll write this kind of a tool in both proposed cases.

If `google-java-format` will be able to traverse over a tree I'll call its "native method" from the script instead of applying it, google-java-format, to the each file. Because it is a more convenient way to use it as a separate tool.

Consider following use case:
```
$ git clone some/java/project.git
$ jfmt
```
Now you are able to read formatted code like it was formatted from the beginning. 

Or
```
$ jfmt
$ git commit -am "comment"
```
To write code in comfortable style and format it afterwards to common style. I can't say that last example can be considered as a good one, but it shows the point.

Cheers.

cushon:
Thanks for the patch! I'm not sure it's worth adding a separate wrapper for this, though. Does `google-java-format -i $(find . -type f -name "*.java")` do what you need?

@sherter good idea, I filed #54 to separate the CLI into its own artifact.

pvlbzn:
Thanks for answer @cushon !

This script is aimed on an integration into users shell. The key point is to type 4 characters and format whole project at once, or type 5 characters and drag-n-drop particular `.java` file into the shell to format it. IDE environments usually have a built-in terminal feature, so it can used out of one programming env.

However, perhaps it will be too much to have two wrappers. I proposed this tool because it serves well to me, with an unix-like idea "one tool for a one job". May be it is a good idea to add this functionality into the existing script? Otherwise lets close this request.

Cheers!

PR #53, opened 2016-06-02
Support --aosp flag in google-java-format-diff.py

benol:
If called with --aosp, pass the flag to the google-java-format binary.


cushon:
Looks good, thanks.

PR #57, opened 2016-07-13
Expose max length to JavaFormatterOptions builder

talios:
For my googleformatter-maven-plugin I allow for customised line lengths, which is no longer possible with the current SNAPSHOT API ( I used to have a subclass that overloaded the `getMaxLineLength()` method.

The `JavaFormatterOptions` class lists a `DEFAULT_MAX_LINE_LENGTH` but it's not actually customisable.

This patch simply exposes that to the `Builder`.

NB: I've not currently signed any CLA's for Google - what ( if any ) is the process there?

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/JavaFormatterOptions.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/JavaFormatterOptions.java
@@ -62,13 +62,16 @@ public class JavaFormatterOptions {
 
   private final Style style;
 
-  private JavaFormatterOptions(Style style) {
+  private final int maxLineLength;
+
+  private JavaFormatterOptions(Style style, int maxLineLength) {
     this.style = style;
+    this.maxLineLength = maxLineLength;
   }
 
   /** Returns the maximum formatted width */
   public int maxLineLength() {
-    return DEFAULT_MAX_LINE_LENGTH;
+    return maxLineLength;
   }
 
   /** Returns the multiplier for the unit of indent */
@@ -89,6 +92,7 @@ public class JavaFormatterOptions {
   /** A builder for {@link JavaFormatterOptions}. */
   public static class Builder {
     private Style style = Style.GOOGLE;
+    private int maxLineLength = DEFAULT_MAX_LINE_LENGTH;
 
     private Builder() {}
 
@@ -97,8 +101,13 @@ public class JavaFormatterOptions {
       return this;
     }
 
+    public Builder maxLineLength(int maxLineLength) {
+      this.maxLineLength = maxLineLength;
+      return this;
+    }
+
     public JavaFormatterOptions build() {
-      return new JavaFormatterOptions(style);
+      return new JavaFormatterOptions(style, maxLineLength);
     }
   }
 }


talios:
Updated my CLA details with my github account.

talios:
I signed it! Noticed my other CLA was specifically an "Android" one, not that I've done any android submissions - odd.

talios:
Any chance of a merge? Would be good to start tracking upstream again.

sherter:
This probably applies here:
> There is no configurability as to the formatter's algorithm for formatting. This is a deliberate design decision to unify our code formatting on a single format.

I guess making this functionality public was a mistake in the first place and they try to correct it now. However, I agree  that it is kind of rude to just remove things from a public API after the 1.0 release...

talios:
> There is no configurability as to the formatter's algorithm for formatting. 

I realise there's the existing no-config ruling, altho I'd argue that we're not exposing the ability change the rules of formatting per sae - such as where to wrap, spacing, indents etc. only the ability to customize the maximum line length.

This change doesn't change the predefined `Style` for `GOOGLE` or `AOSP`, only retains the existing functionality for flavours built with the builder.

I find that when doing things with a lot of generics, 100 chars isn't quite wide enough ( but then I find the formatting also yells "oi, refactor this!" ).

Perfectly understandable if we want to remove the ability to config width totally tho.

talios:
Any further decision on this?

talios:
No further comments on this? I'd still like to retain the ability to change/set the max-width of formatting, which isn't really changing any of the formatting algorithms - i.e. line wrapping rules, indentation rules.

It's just allowing to work, what originally worked.

talios:
Congrats on the 1.1 release, any change this can be pulled in for the next release? Or closed....

fzakaria:
+1 I'd love to see this merged.

talios:
@fzakaria You and me both, for now I've removed this functionality from my maven-google-formatter plugin tho :(

kd8azz:
I understand the motivation behind not allowing configurability.  But the notion of a flow-layout is extremely prevelant in UI design, and line wrapping is part of that.  In a perfect world, code would be saved to disk with a standard line length, but displayed in my IDE with a line length matching the width of my window.

Adding a line length option to the formatter would allow IDE plugins to be built to add this functionality, without sacrificing the standards upheld by the no-configurability rule.

Please reconsider.

DwordPtr:
We really need a configurable code-formatting tool in the java community. Trying to get Jalopy to work is just painful. This could be done in a very non-invasive way like a properties file or even static constants.

I understand google is trying to maintain a unified formatter but how about leaving a side door open for those of us in other environments.

russgold:
If there has to be a fixed length, at least make it larger, say 132. 100 is way too short, and encourages the use of short (and therefore difficult to read) variable names.

kevinb9n:
Everyone, I'm very sorry for our extended inattention to this request. Of course, it came from us being internally conflicted, but I meant to actually bring that discussion *here*.

So, there's a few layers here. The first is why our general philosophy against configurability is so deeply held for our internal development.

You've noticed of course that we expose extremely few configuration options that govern formatting behavior. We have a few options that are used only to switch entire behaviors on or off, and we have --aosp (which internally to our company we don't even surface in the various tool integrations). We've held the line at just those; why?

The explicit goals we had for building this tool are to bring consistency to Java code, and to free developers from arguments over style choices. It has always been an explicit non-goal to support developers' individual preferences, because doing so actually works directly against the primary goals.

Then it turned out that our lack of configuration also enabled us to deliver a much more thoroughly-vetted, high-quality tool. As configurable parameters rise, the complexity of testing rises exponentially. To us this is not just a theoretical benefit.

Finally, sometimes your preferred formatting is not merely a matter of taste, but one that might be just plain superior to the current formatter behavior. In that case, we *want* to force a conversation about which behavior is really best, so that everyone can benefit. Configurability would sidestep all that. (I'll hasten to point out that this particular argument does not apply to line length, because in that case we are quite strongly committed to the choice of 100 that we settled on many years ago, and believe that no good can come from relitigating that number every few years.)

We fully understand that not everyone wants to accept such an inflexible formatter as this, but we have figured that such users would probably not choose google-java-format.

Is the line length issue, in particular, special in some other way? One thing I thought of is that I might be formatting a piece of code to display in a document/slideshow/code-sample-embedded-in-javadoc/etc. I would want to set a narrower limit in these cases. If this is a use case that people really have, that is actually a little more convincing to us than that different people want their checked-in code to look differently.

Hope this at least helps the conversation move forward. Again, I apologize for our silence.

tbroyer:
> Is the line length issue, in particular, special in some other way? One thing I thought of is that I might be formatting a piece of code to display in a document/slideshow/code-sample-embedded-in-javadoc/etc. I would want to set a narrower limit in these cases. If this is a use case that people really have, that is actually a little more convincing to us than that different people want their checked-in code to look differently.

Fwiw, for those kind of one-off use-cases, you might want to use a different formatter with configurable line length; e.g. https://github.com/thorbenvh8/prettier-java

As for configurable Java code formatters, well, there's Eclipse JDT. If you're using Gradle or Maven, it can be integrated through [Spotless](https://github.com/diffplug/spotless)

alexkleiman:
@tbroyer, thank you for the detailed and well-considered response. Your thinking makes a lot of sense, and although your decision isn't what we were hoping for, it seems well thought out and reasonable.

To address a specific question you raise: "Is the line length issue, in particular, special in some other way?"

The main reason why I see line length as special is a purely practical consideration: it is much more likely to block the adoption of google-java-format than many other formatting decisions. When I reformatted my organization's Java codebase, the only issue people were concerned about was the line length. Because changing our maximum line length would have antagonized so many people, the only viable option was for me to use a fork of google-java-format which allowed us to use a different line length.

Another practical reason why line length is special is that changing the line length of an existing codebase can be difficult. Normally when line-wrapping and continuations get too hard to read, some of the code can just be broken out into a helper method. When reformatting an existing codebase, it's hard to write dozens or hundreds of new helper methods to avoid the use of too many continuations.

Due to these two considerations, I believe that the lack of a configurable line length is a more significant barrier to adoption of `google-java-format` than is any other formatting decision the formatter makes.

Turning to the points you've raised, it seems like you've outlined three different advantages to not having a configurable line length:

1) "The explicit goals we had for building this tool are to bring consistency to Java code, and to free developers from arguments over style choices"
2) "As configurable parameters rise, the complexity of testing rises exponentially"
3) "we want to force a conversation about which behavior is really best, so that everyone can benefit. Configurability would sidestep all that."

If people who need a line length of other than 100 simply use forks of `google-java-format` or turn to other formatters, then goals (1) and (3) aren't satisfied by keeping line length non-configurable (although if everybody within Google uses `google-java-format, then I suppose (1) and (3) would still apply to a large number of developers).

If (1) and (3) don't impact the community at large, then the biggest tradeoff seems to be between, on one hand, avoiding the need to test and maintain a configurable line length, and on the other hand, increasing adoption of `google-java-format`. If configurable line length is deemed to much trouble to support, I would understand that decision -- such is the nature of software engineering. As an enthusiastic user of `google-java-format`, though, I hope that eventually a time would come when your team is in a position to be able to bear the cost of supporting configurable line length.

Once again, thank you for your careful consideration of the points raised in the discussion of this pull request.

kd8azz:
@kevinb9n To reiterate my position (which is purely theoretical at this point), configurable line length is most attractive to me for displaying, not for checking in.  This matches the point you made about formatting code for a presentation.  But in my case, I would envision an IDE plugin which would display code formatted to the current width of the window, while saving it formatted to 100 characters.

kd8azz:
> In such a future what does it matter what length you you save it in :)
Backwards compatibility with the old standard is important, because other people use different IDEs.

fletchgqc:
@alexkleiman Is anyone maintaining a fork of google-java-format with configurable line length? Where can we get it?

talios:
@fletchgqc sadly not I. And I've not heard of anyone doing so.

talios:
@kd8azz Yeh - take a look at the pull request I added way back in 2016 to _restore_ this functionality - https://github.com/google/google-java-format/pull/57/commits/0dab0dd15ade0f802231c416bbda5e3c1f6bd482

alexkleiman:
@fletchgqc just to echo @talios - unfortunately I don't know of any fork being maintained.

russgold:
For all my complaints about line length - I have been using it to motivate more refactoring. In theory, it would be easy to make a fork, but the architect of my project wouldn’t use it. Having ONE TRUE STANDARD has a value to him. If we had line-length as a configuration, we’d have arguments about what the right length is. This way, we don’t, because it’s not an option.

VitaliyKulikov:
I prefer to stay with 100 and unify settings.  code review goes better without scrolling.

cushon:
To add to what @kevinb9n said earlier, we now have the beginnings of an FAQ that includes some discussion of the approach towards (non-)configurability [here](https://github.com/google/google-java-format/wiki/FAQ#i-just-need-to-configure-it-a-bit-differently-how).

jaytaylor:
FWIW, +1, I'd very much line to have maxLineLength exposed as an option.

Applying a 100 character limit to Java seems archaic in this day and age of high-resolution displays, especially where displays are typically wider than they are tall.

jaytaylor:
If you want the `--max-line-length` flag, I've published a release in my fork (this is only change).

The fat jar is available for download:

https://github.com/jaytaylor/google-java-format/releases

https://github.com/jaytaylor/google-java-format/releases/download/1.8.1/google-java-format-1.8.1-all-deps.jar

Cheers.

talios:
@jaytaylor I can understand their reasoning in part for documentation/screenshot/slideshow reasons, but wide screens and java generics don't really help much there - maybe once we can all use `var` it might help.

My maven plugin has a....  hacky workaround which restores this capability via the plugin tho.

jaytaylor:
@talios I understand the reasoning too, it's pretty annoying to contrast this reasoning against golang's `gofmt`, which does not enforce any line length limitations.

All I hope is that I can save someone else from having to go through and find a way to add this flag :)

talios:
I see the PR is now conflicting due the magic API breaking change of actually removing the underlying line settings in the JavaFormatterOptions class - however this breaking change doesn't seem to have had a major version number change associated with it ( moving to 2.x ).

This will break integration with at least my maven plugin without other changes, I'm not sure about the others.

timrobertson100:
Another +1 here

A default of 100, but some hook that allows others to use something else seems reasonable. In my experience, it is sometimes the *only* thing that blocks some people from using this.

talios:
I had worked around this in my plugin by using Mockito and, rather nastily using a `spy()` to mock the length to configured length, which worked well. But since the API has now been physically removed I don't be able to do so.

However, that's largely moot as GCF as now moved to a >JDK8 requirements, which means I can't use it in any Java 8 based builds - which I'm stuck with for the current time, so I doubt I'll be updating my plugin.

natros:
Don’t use the plugin. Is it’s slower to format big files. What I did was building the project and generated a native executable (graalvm) that I’m using with File Watchers. Every time the file is saved or compiled the file is formated.


Filipe Sousa

jeffkreska:
I think these are some points about wrapping that should be considered before hardcoding a wrapping length

1. All modern editors support auto wrapping
2. You don't know what size screen is viewing the code so 100 does not fit for everyone (i.e. mobile phone, tablet, ...)
3. Wrapping was created to enable printouts to be readable in the 1980's when editors/software was not smart enough to wrap for the display or printer.
4. Wrapping has nothing to do with a code style (point made above years ago)
5. Hardcoding a line wrap width does not enforce a common code style, it does force everyone to use an editor that shows at least 100 characters or the double wrapping (fixed 100 char wrap and editor builtin wrapping when viewing in less than 100 char wide display)

Maybe I am to old fashioned but I believe it is an anti-pattern to assume you know what everyone else wants/needs.

owlg:
Seriously, this is not yet approved?

jbduncan:
@owlg Realistically, it's unlikely to ever be approved, because the GJF team want to minimise the number of configuration options so that teams using GJD don't have to argue over code style as much.

tbroyer:
* https://github.com/google/google-java-format/pull/57#issuecomment-388125762
* https://github.com/google/google-java-format/wiki/FAQ#i-just-need-to-configure-it-a-bit-differently-how

PR #77, opened 2016-09-29
Add support for javac-style @params files

sormuras:
This PR adds support for javac-style @params files.

Example: `@params.txt`:
```
--dry-run
--set-exit-if-changed
@@-escaped-option-starting-with-a-single-@-character

α.java
β.java
Ω.java

@nested-params-file
```

This solves two problems with current command line program:
- very long lists of Java source files are cluttering the command line
- there's Windows-related bug, when Unicode encoded file names are passed via the command line arguments. The non-Unicode named `params.txt` may contain Unicode source file names. See https://stackoverflow.com/questions/7660651/passing-command-line-unicode-argument-to-java-code

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
@@ -14,9 +14,19 @@
 
 package com.google.googlejavaformat.java;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import com.google.common.base.CharMatcher;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableRangeSet;
 import com.google.common.collect.Range;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
@@ -25,11 +35,13 @@ final class CommandLineOptionsParser {
 
   private static final Splitter COMMA_SPLITTER = Splitter.on(',');
   private static final Splitter COLON_SPLITTER = Splitter.on(':');
+  private static final Splitter ARG_SPLITTER =
+      Splitter.on(CharMatcher.breakingWhitespace()).omitEmptyStrings().trimResults();
 
   /** Parses {@link CommandLineOptions}. */
   static CommandLineOptions parse(Iterable<String> options) {
     CommandLineOptions.Builder optionsBuilder = CommandLineOptions.builder();
-    Iterator<String> it = options.iterator();
+    Iterator<String> it = expandParamsFiles(options).iterator();
     while (it.hasNext()) {
       String option = it.next();
       if (!option.startsWith("-")) {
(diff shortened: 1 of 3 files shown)


lowasser:
If I were doing this on a Linux box I'd just write `cat file | xargs run-formatter`.  Is there not something equivalent in your environment?

sormuras:
Maybe. I think, not.

The main problem is the command line / shell environment on Windows. If those programs you mention exists on Windows, they'll still use the non-Unicode ... thing. If I'm not missing something, it would still not work.

See the `MainTest#testMainWithUnicodeArgumentFailsOnWindows` included in this PR:
```java
    String expected = "ℕ.java";
    String actual = "N.java"
```
And this is a good (human-readable) guess -- almost all other Unicode characters are translated to `?`s.

core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
```diff
@@ -99,6 +110,25 @@ static CommandLineOptions parse(Iterable<String> options) {
     return optionsBuilder.build();
   }
 
+  private static void parseSourcesFile(ImmutableList.Builder<String> filesBuilder, String value) {
+    Path path = Paths.get(value);
+    try {
+      for (String line : Files.readAllLines(path, StandardCharsets.UTF_8)) {
+        line = line.trim();
```
jbduncan: I wonder if this line would be less bug-prone if it were replaced with `CharMatcher.whitespace().trimFrom(line)`.

sormuras: Replaced.

cushon:
What do you think about supporting javac-style `@params` files, instead of doing this specifically for the sources argument?

sormuras:
In progress...

cushon:
Maybe something like this:
https://github.com/google/turbine/blob/a5453051f0c8b71a833f6dc73caaea43bd24a0db/java/com/google/turbine/options/TurbineOptionsParser.java#L62

sormuras:
Ah, with escaping. Check.

sormuras:
Expand the "args" in `Main#processArgs(String...)` or `CommandLineOptionsParser#parse(Iterable<String>)`?

I go with the second option for an initial PR.

sormuras:
Ready for review.

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -119,6 +119,37 @@ public void testMain() throws Exception {
     assertThat(process.exitValue()).isEqualTo(0);
   }
 
+  @Test
+  // https://stackoverflow.com/questions/7660651/passing-command-line-unicode-argument-to-java-code
+  public void testMainWithUnicodeArgumentFailsOnWindows() throws Exception {
```
cushon: Did you want to include this test, or was it more to document the problem? I guess it would provide some notice if this changed?

core/src/test/java/com/google/googlejavaformat/java/CommandLineOptionsParserTest.java
```diff
@@ -152,4 +160,12 @@ public void illegalLines() {
...
   }
 
+  @Test
+  public void sourcesFile() throws IOException {
+    Path path = testFolder.newFile("HelloAndGoodbye.txt").toPath();
+    Files.write(path, "Hello.java\n   \n\nℕ.java\nGoodbye.java".getBytes(UTF_8));
+    CommandLineOptions options =
+        CommandLineOptionsParser.parse(Arrays.asList("@" + path.toString()));
+    assertThat(options.files()).containsExactly("Hello.java", "ℕ.java", "Goodbye.java");
+  }
```
cushon: Can you also test non-params file arguments before/after the params file, the `@@`-escape syntax, and maybe a nested params file?

sormuras: Sure.

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
sormuras: It only demonstrates the underlying issue why I needed that `--sourceFile` feature.

I think, it's superseded by now and I'll remove it from this PR.

sormuras:
Should also support comment lines? Meaning ignore lines that start with `//`.

sormuras:
Almost down to a [single feature branch](https://github.com/sormuras/google-java-format/branches). :)

Issue #33, opened 2016-03-25
Windows support


cushon:
The formatter makes some assumptions that don't work on Windows.

- [x] `File.renameTo()` is not portable (see #32)
- [x] The output always uses `\n` as a line terminator, we should probably use the platform terminator

sormuras:
Two tests in `DiagnosticTest` fail on Windows due to hard-coded line separators in combination with `java.io.PrintWriter` dynamic behaviour:
```java
    /**
     * Line separator string.  This is the value of the line.separator
     * property at the moment that the stream was created.
     */
    private final String lineSeparator;
```
How do you want it to be fixed?
 * remove `\n` from end of expected string and use `contains` instead of `isEqualTo`

`assertThat(err.toString()).contains("<stdin>:2:4: error: Invalid character constant");` 

 * replace `\n` with `System.lineSeparator()`

`assertThat(err.toString()).isEqualTo("<stdin>:2:4: error: Invalid character constant" + System.lineSeparator());`?

jbduncan:
I recently suggested in junit-team/junit5#398 the use of AppVeyor to prevent these sorts of Windows-related errors from cropping up in the future.

What does the google-java-format team think about using AppVeyor here?

sormuras:
AppVeyor would report: `Tests run: 509, Failures: 270, Errors: 0, Skipped: 1` ... all `CRLF` related.

jbduncan:
Haha, well then! Looks like a pretty convincing case for using AppVeyor to me.

sormuras:
After converting all `.[in|out]put` files to `\n` line separators, the test reports: `Tests run: 507, Failures: 15, Errors: 0, Skipped: 1`

Maybe a `.gitattributes` with:
```
*.input -text
*.output -text
```
The remaining 15 failures are extra `\n` characters where none are expected. Like in `MainTest.importRemovalLines`, there's an empty line rendered between the `import` and `class` block.

sormuras:
#79 proposes a dynamic way for the line separator issue, which does not override the user choice of which line separator to use.

May be, another command line option can be introduced using an enum named `LineSeparator`:
- AUTO_DETECT // use the separator used in input
- DEFAULT // use the system-dependend default separator via System.lineSeparator()
- UNIX // "\n"
- MAC // "\r"
- WINDOWS / "\r\n"

sormuras:
[...]

PR #79, opened 2016-10-05
Auto detect line separator from input and use it in output.

sormuras:
# Overview
Solves #33 by auto detecting line separator from input, potentially convert to U`\n`ix-style, format/fix/remove-used and revert conversion before output, if necessary.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/LineSeparator.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/LineSeparator.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.googlejavaformat.java;
+
+import java.util.regex.Pattern;
+
+/**
+ * Line separator constants and utilities.
+ */
+public enum LineSeparator {
+
+  /**
+   * Windows line separator: CRLF or as String {@code "\r\n"}.
+   */
+  WINDOWS("\r\n", Pattern.compile("(?s).*(\\r\\n).*")),
+
+  /**
+   * Unix/Linux line separator: LF or as a String {@code "\n"}.
+   */
+  UNIX("\n", Pattern.compile("(?s).*(\\n).*")),
+
+  /**
+   * Legacy Mac OS line separator: CR or as a String {@code "\r"}.
+   */
+  MAC("\r", Pattern.compile("(?s).*(\\r).*"));
+
+  /**
+   * Convenient for {@code detect(text, LineSeparator.UNIX)}.
+   */
+  public static LineSeparator detect(String text) {
+    return detect(text, UNIX);
+  }
+
+  /**
+   * Identify which line delimiter is used by the underlying text.
+   *
+   * <p>The enum constants are iterated in declaration order by this implementation. That means for
+   * texts using mixed separators, the first hit determines the returned separator of this method.
+   *
+   * <p>If no pattern matches, the passed default separator is returned.
+   */
+  public static LineSeparator detect(String text, LineSeparator defaultSeparator) {
+    for (LineSeparator separator : values()) {
+      if (separator.pattern.matcher(text).matches()) {
+        return separator;
+      }
+    }
+    return defaultSeparator;
+  }
+
+  private final String chars;
+  private final Pattern pattern;
+
+  LineSeparator(String chars, Pattern pattern) {
+    this.chars = chars;
+    this.pattern = pattern;
+  }
+
+  public String getChars() {
+    return chars;
+  }
+
+  /**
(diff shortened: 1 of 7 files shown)


sormuras:
Finally w/o failures.

core/src/main/java/com/google/googlejavaformat/java/LineSeparator.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
jbduncan: Minor nit: this should read `Copyright 2016 Google Inc.` for consistency with the other source files.

core/src/main/java/com/google/googlejavaformat/java/LineSeparator.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.regex.Pattern;
+
+/** Line separator constants and utilities. */
+public enum LineSeparator {
+   WINDOWS("\r\n"),
+
+   UNIX("\n"),
+
+   MAC("\r");
```
jbduncan: Another nit: even though at least one enum class e.g. `JavaFormatterOptions.Style` separates enums with additional lines, I wonder if in this case they'd read better without them. So something like

```java
WINDOWS("\r\n"),
UNIX("\n"),
MAC("\r");
```

core/src/test/java/com/google/googlejavaformat/java/FormatterTest.java
```diff
@@ -309,4 +312,16 @@ public void stringEscapeLength() throws Exception {
     assertThat(new Formatter().formatSource("class T {{ f(\"\\\"\"); }}"))
         .isEqualTo("class T {\n  {\n    f(\"\\\"\");\n  }\n}\n");
   }
+
+  @Test
+  public void nonUnixStyleFails() throws Exception {
+    String win = "class T {\r\n}\r\n";
+    try {
+      new Formatter().formatSource(win, Collections.singleton(Range.closedOpen(0, win.length())));
+      fail("Windows-style line separator should fail!");
```
jbduncan: Another nit: that exclamation mark looks a bit...  sudden! 😉

Could it be replaced with a full stop?

core/src/main/java/com/google/googlejavaformat/java/LineSeparator.java
sormuras: Copied it from `RemoveUnusedImports.java`. Will use another source for copy & paste.

core/src/test/java/com/google/googlejavaformat/java/FormatterTest.java
sormuras: It should never be visible, anyway. Sudden feels right, imho.

core/src/main/java/com/google/googlejavaformat/java/LineSeparator.java
sormuras: Added some meaningful javadoc comment, now the constant block looks better when separated. :wink:

cushon:
Thanks. I think we'd like to avoid the preprocessing eventually and have CRLF/LF/CR all treated the same, but that's going to be more work and this seems like a good first step.

I think the current approach breaks partial formatting using `-offset` and `-length`, though?

```
echo -e "class Test {\r\n\r\n}" > Test.java
google-java-format -offset 16 -length 1 Test.java
Test.java:error: invalid length 1, offset + length (17) is outside the file
```

Also, are your thoughts on using `System.lineSeparator()` for output vs. preserving the separator in the input?

sormuras:
> ...have CRLF/LF/CR all treated the same...

That was my initial approach, too. The internal processing of line breaks is really based on `\n`. If it were only the emitting part, but there're many scanning/processing switches depending on that single char.

> ...current approach breaks partial formatting...

No unit test is complaining. I'll add your test case. :wink:

> ...'System.lineSeparator()' vs. preserving the separator in the input?

I think, that they are two different concerns. Like many IDEs do separate them (`File - Convert Line Terminators` vs. `Code - Reformat`), I'd expect a formatting tool to leave the current line separators alone. Until I state/force otherwise. That's where a "second" command line switch, I proposed at #33 might help.

sormuras:
I fixed the partial formatting issue by correcting the offset and length values on the fly - if necessary.

sormuras:
@cushon I won't be able to resolve the conflict until Monday.

sormuras:
Merged and rebased. Anything else to do?

cushon:
It'd like to avoid the preprocessing - it's not free, and the difference between the entry points could be surprising. I took another look at recognizing all all input line endings and came up with 4174d878b36159a8efec949e95531d6cc1276393. Can you see if that works for you?

It still uses '\n' for output, so that hasn't been fixed yet.

sormuras:
Formatting CRLF files seems, with the `\n` used for output exception, ok. 👍 

When implementing the output part, please consider leaving the line separator "as is". See comment https://github.com/google/google-java-format/issues/33#issuecomment-251667538 for details. Their are projects, that use `.gitattributes` or other measures, keeping the line seperator constant -- no matter what the client platform uses as default.

sormuras:
Shall I file an output PR against current master or do you already have something in store?

cushon:
> Shall I file an output PR against current master or do you already have something in store?

I haven't started anything, so I'd be happy to take a PR for that. I'd prefer to change the code that writes newlines instead of adding a post-processing pass, though. And I didn't realize there was demand for leaving newlines as-is, so doing that instead of just using `System.lineSeparator()` seems OK.

sormuras:
I'm almost done - only 15 extra-newlines within import/javadoc ranges left.

sormuras:
Superseded by #87

jom0000:
ควย

PR #88, opened 2016-10-19
Fix test failures on Windows

sormuras:
Failures in `MainTest` are easy -- [PrintWriter](http://docs.oracle.com/javase/7/docs/api/java/io/PrintWriter.html)s methods use the platform's own notion of line separator rather than the newline character. Just use the system default separator when compiling the expected output.

~~Failures in `RemoveUnusedImportsTest` are **green** on Windows now, but I can't trace an use of `PrintWriter` here. Ah... there's ringing a bell: didn't the `RemoveUnusedImports` somewhere refer to `System.lineSeparator`?~~ BINGO.

Force-pushing the real fix in a minute to this branch.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/RemoveUnusedImports.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/RemoveUnusedImports.java
@@ -25,6 +25,7 @@ import com.google.common.collect.RangeMap;
 import com.google.common.collect.RangeSet;
 import com.google.common.collect.TreeRangeMap;
 import com.google.common.collect.TreeRangeSet;
+import com.google.googlejavaformat.Newlines;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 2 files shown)


cushon:
> PrintWriters methods use the platform's own notion of line separator rather than the newline character.

Sorry - I didn't realize what this change was for in the original PR. I'd rather not have the tests depend on the platform line separator, but I forgot about PrintWriter.

> didn't the RemoveUnusedImports somewhere refer to System.lineSeparator?

Oh, right. Does using `guessLineSeparator` work? https://github.com/google/google-java-format/blob/e345bce16276a6c0d8e50ba184b9ae1e02e71478/core/src/main/java/com/google/googlejavaformat/java/RemoveUnusedImports.java#L212

sormuras:
Ja, already pushed the better fix.

PR #90, opened 2016-10-27
Refactored to interface, version constant updated.

sormuras:
Made type `GoogleJavaFormatVersion` an interface - not instantiable. Removed redundant modifiers and updated the version string reflect current value used in main `pom.xml`.


cushon:
Oops, the edit to `GoogleJavaFormatVersion` in 13f608bbd6f3f26c3eec05807b3a37bef9bb7670 was an accident.

sormuras:
Never mind. (-:

PR #92, opened 2016-11-02
Added format and optimize imports entry point.

sormuras:
This PR introduces a new entry point in `Formatter`, that format an input string (a Java compilation unit) and optimize imports into an output string.

The new entry point `formatSource(String, JavadocOnlyImports )` behaves similar to the `Main`/`FormatFileCallable` combo, that allows optimizing imports in Java source files. It provides a better manner for external application to invoke the expected behaviour of google-java-format programmatically -- without utilizing the `Main`/`FormatFileCallable` combo path discussed here https://github.com/diffplug/spotless/issues/48

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Formatter.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Formatter.java
@@ -32,6 +32,7 @@ import com.google.googlejavaformat.FormattingError;
 import com.google.googlejavaformat.Newlines;
 import com.google.googlejavaformat.Op;
 import com.google.googlejavaformat.OpsBuilder;
+import com.google.googlejavaformat.java.RemoveUnusedImports.JavadocOnlyImports;
 import com.sun.tools.javac.file.JavacFileManager;
 import com.sun.tools.javac.parser.JavacParser;
 import com.sun.tools.javac.parser.ParserFactory;
@@ -194,12 +195,33 @@ public final class Formatter {
   /**
    * Format an input string (a Java compilation unit) into an output string.
    *
+   * <p>Leaves import statements untouched.
+   *
    * @param input the input string
    * @return the output string
    * @throws FormatterException if the input string cannot be parsed
    */
   public String formatSource(String input) throws FormatterException {
     return formatSource(input, Collections.singleton(Range.closedOpen(0, input.length())));
+  }
+
+  /**
+   * Format an input string (a Java compilation unit) and optimize imports into an output string.
+   *
+   * <p>Optimizing imports includes ordering, spacing and removal of unused import statements. An
+   * import used only in a Javadoc comment is considered as unused and therefore removed from the
+   * import statements and its usage within the comment replaced by its fully qualified name.
+   *
+   * @param input the input string
+   * @return the output string
+   * @throws FormatterException if the input string cannot be parsed
+   * @see <a href="https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing">
+   *   Google Java Style Guide - 3.3.3 Import ordering and spacing</a>
+   */
+  public String formatSourceAndOptimizeImports(String input) throws FormatterException {
+    input = ImportOrderer.reorderImports(input);
+    input = RemoveUnusedImports.removeUnusedImports(input, JavadocOnlyImports.REMOVE);
+    return formatSource(input);
   }
 
   /**
--- a/core/src/test/java/com/google/googlejavaformat/java/FormatterTest.java
+++ b/core/src/test/java/com/google/googlejavaformat/java/FormatterTest.java
@@ -249,6 +249,15 @@ public final class FormatterTest {
   }
 
   @Test
+  public void importsOptimizedIfRequested() throws FormatterException {
+    String input =
+      "package com.google.example;\n" + UNORDERED_IMPORTS + "\npublic class ExampleTest {}\n";
+    String output = new Formatter().formatSourceAndOptimizeImports(input);
+    String expect = "package com.google.example;\n\npublic class ExampleTest {}\n";
+    assertThat(output).isEqualTo(expect);
+  }
+
+  @Test
   public void importOrderingWithoutFormatting() throws IOException, UsageException {
     importOrdering(
         "--fix-imports-only", "com/google/googlejavaformat/java/testimports/A.imports-only");
--- a/core/src/test/java/com/google/googlejavaformat/java/MainTest.java
+++ b/core/src/test/java/com/google/googlejavaformat/java/MainTest.java
@@ -30,6 +30,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.attribute.PosixFilePermission;
 import java.util.EnumSet;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -200,6 +201,40 @@ public class MainTest {
       assertThat(
               main.format("-", "--fix-imports-only", "--experimental-remove-javadoc-only-imports"))
           .isEqualTo(0);
+      assertThat(out.toString()).isEqualTo(joiner.join(expected));
+    }
+  }
+
+  @Test
+  @Ignore("FormatFileCallable.call() needs a second format run after fixing imports")
+  public void optimizeImportsDoesNotLeaveEmptyLines() throws Exception {
+    String[] input = {
+      "package abc;",
+      "",
+      "import java.util.LinkedList;",
+      "import java.util.List;",
+      "import java.util.ArrayList;",
+      "",
+      "import static java.nio.charset.StandardCharsets.UTF_8;",
+      "",
+      "import java.util.EnumSet;",
+      "",
+      "class Test ",
+      "extends ArrayList {",
+      "}"
+    };
+    {
+      String[] expected = {
+        "package abc;", "", "import java.util.ArrayList;", "", "class Test extends ArrayList {}", ""
+      };
+      // pre-check expectation with local formatter instance
+      String optimized = new Formatter().formatSourceAndOptimizeImports(joiner.join(input));
+      assertThat(optimized).isEqualTo(joiner.join(expected));
+
+      InputStream in = new ByteArrayInputStream(joiner.join(input).getBytes(UTF_8));
+      StringWriter out = new StringWriter();
+      Main main = new Main(new PrintWriter(out, true), new PrintWriter(System.err, true), in);
+      assertThat(main.format("-")).isEqualTo(0);
       assertThat(out.toString()).isEqualTo(joiner.join(expected));
     }
   }


sormuras:
@cushon Any chance to merge this or a similar entry point?

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -203,6 +204,25 @@ public String formatSource(String input) throws FormatterException {
...
 
   /**
+   * Format an input string (a Java compilation unit) and optimize imports into an output string.
+   *
+   * @param input the input string
+   * @param joi if not {@code null}, the imports are optimized accordingly
+   * @return the output string
+   * @throws FormatterException if the input string cannot be parsed
+   */
+  public String formatSource(String input, JavadocOnlyImports joi) throws FormatterException {
```
cushon: `JavadocOnlyImports` is an experiment, let's leave it out of the API for now.

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -203,6 +204,25 @@ public String formatSource(String input) throws FormatterException {
...
+    String output = formatSource(input, Collections.singleton(Range.closedOpen(0, input.length())));
```
cushon: I think you can use `formatSource(String)` instead?

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -203,6 +204,25 @@ public String formatSource(String input) throws FormatterException {
...
+    if (joi != null) {
+      output = RemoveUnusedImports.removeUnusedImports(output, joi);
+      output = ImportOrderer.reorderImports(output);
+      // re-run format to remove empty lines and other artifacts introduced by imports mangling...
```
cushon: What other artifacts? Is there a bug for that? If there are issues with the import pass, let's fix that instead of formatting everything twice.

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
sormuras: Tried that first, but there were some test failing - will list them here.

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
sormuras: Multiple empty lines between `package` and `class` declaration are left behind by the import pass.

sormuras: Instead of

    package com.google.example;
     
    public class ExampleTest {}

it produces

    package com.google.example;
    
    
    
    
    
    
    public class ExampleTest {}

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
sormuras:
```
org.junit.ComparisonFailure: bad output for B21647014  <Click to see difference>
org.junit.ComparisonFailure: bad output for B21647014  <Click to see difference>
org.junit.ComparisonFailure: bad output for B21647014  <Click to see difference>
org.junit.ComparisonFailure: bad output for B21647014  <Click to see difference>

org.junit.ComparisonFailure: bad output for M  <Click to see difference>
org.junit.ComparisonFailure: bad output for M  <Click to see difference>
org.junit.ComparisonFailure: bad output for M  <Click to see difference>
org.junit.ComparisonFailure: bad output for M  <Click to see difference>

org.junit.ComparisonFailure: bad output for Unformatted  <Click to see difference>
org.junit.ComparisonFailure: bad output for Unformatted  <Click to see difference>
org.junit.ComparisonFailure: bad output for Unformatted  <Click to see difference>
org.junit.ComparisonFailure: bad output for Unformatted  <Click to see difference>

org.junit.ComparisonFailure:  <Click to see difference>
	at com.google.common.truth.Platform.comparisonFailure(Platform.java:133)
	at com.google.common.truth.Truth$1.failComparing(Truth.java:77)
	at com.google.common.truth.StringSubject.isEqualTo(StringSubject.java:69)
	at com.google.googlejavaformat.java.FormatterTest.importsNotReorderedByDefault(FormatterTest.java:249)


org.junit.ComparisonFailure:  <Click to see difference>
	at com.google.googlejavaformat.java.JavadocFormattingTest.doFormatTest(JavadocFormattingTest.java:1269)
	at com.google.googlejavaformat.java.JavadocFormattingTest.brAtSignBug(JavadocFormattingTest.java:823)
```
The initial blocks of four failures are from `FormatterIntegrationTest.idempotent...` parameterized test.

I vote for keeping the new method name for the new behaviour.

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
sormuras: What about keeping the extra pass with a `TODO` marker, open a new ticket for it - and solving it afterwards?

sormuras: Found the reason: in `private static JCCompilationUnit parse(Context context, String javaInput)` the `javaInput` String provided by the test uses `\n` as line separator - but the `JCCompilationUnit` uses `\r\n`, presumably the system default line separator.

Do you know off-hand, how pass our guessed line separator to the context, parser, thing from `com.sun.source.WhatEver`?

sormuras: Or do it like:

1. reorder imports (at most 2 blocks of import statements)
2. remove unused imports (at most 2 empty lines above the first type declaration)
3. format

sormuras:
@cushon I refactored the entry point to `Formatter.formatSourceAndOptimizeImports(String)`, leaving `JavadocOnlyImports` out of the API. Concerning the method name: if we re-use the already published `formatSource(String)` some test fail. See details in the outdated comments.

I added an ignored test to `MainTest`:
```java
  @Test
  @Ignore("FormatFileCallable.call() needs a second format run after fixing imports")
  public void optimizeImportsDoesNotLeaveEmptyLines()
  ...
```
that demonstrates the _artifacts_ when format is executed before fixing imports. I tried to re-order the `call()` method as:
```java
@Override
public String call() throws FormatterException {
  String fixed = fixImports(input);
  if (parameters.fixImportsOnly()) {
    return fixed;
  }

  Formatter formatter = new Formatter(options);
  return formatter.formatSource(fixed, characterRanges(fixed).asRanges());
}
```
but that leads to many other failing assertion, mainly error message checking ... new issue?

sormuras:
Anything left to do, @cushon?

sormuras:
@cushon Still interested in merging this? If yes, I'll resolve the conflicts in `Formatter.java`

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -32,6 +32,7 @@
 import com.google.googlejavaformat.Newlines;
 import com.google.googlejavaformat.Op;
 import com.google.googlejavaformat.OpsBuilder;
+import com.google.googlejavaformat.java.RemoveUnusedImports.JavadocOnlyImports;
```
jbduncan: @sormuras Is this import needed any longer?

core/src/test/java/com/google/googlejavaformat/java/FormatterTest.java
```diff
@@ -249,6 +249,15 @@ public void importsNotReorderedByDefault() throws FormatterException {
   }
 
   @Test
+  public void importsOptimizedIfRequested() throws FormatterException {
+    String input =
+      "package com.google.example;\n" + UNORDERED_IMPORTS + "\npublic class ExampleTest {}\n";
```
jbduncan: @sormuras Does not Google Java Style dictate that indents in multiline statements must be +4, or am I missing something here?

cgrushko:
Any update on this?
I was surprised to find out that the formatter behaves differently on the command-line and in the IntelliJ plugin.

cushon:
+@plumpy

> I was surprised to find out that the formatter behaves differently on the command-line and in the IntelliJ plugin.

The IntelliJ plugin leaves import sorting and unused import removal to IntelliJ. The import fixing operates on the entire file instead of only handling changes lines, and when we did the Eclipse integration it really didn't want the formatter touch anything outside of the specific ranges it requested. Maybe IntelliJ handles that better?

Anyway, IntelliJ's built-in import removal is going to be better integrated and more accurate, since it has a compiler front-end instead of identifying uses of imports syntactically. And it's easy to set up IntelliJ to use the new import order. So I'm not sure we actually want to use GJF's import handling there?

sormuras:
I'd still be happy to resolve the conflicts. :)

cgrushko:
@cushon it's an extra thing to configure (which I didn't know about).
But sure :)
Is there some guide on how to do this? perhaps put it in README.md?

cushon:
If changing the configuration is the best option, documenting that in the README sounds good. But I was hoping @plumpy would have a better idea about how to make it automatic :)

plumpy:
@cushon I can look into it. There is some additional configuration which we do automatically for Google-internal installs of IntelliJ, so that's why I had never really noticed gjf wasn't the one changing the imports. @cgrushko was using a non-Google install at the time, so that's why he did notice :)

I'll see if I can force IntelliJ to use gjf's import style when it's enabled.

opporancisis:
Dear supporters of this project. Is there any way to fix the issue with imports ordering in Intellij IDEA?

Current behavior:
 - install plugin, enable it
 - open any class that previously was in a bad format
 - Ctrl+Alt+L
 = code is formatted as expected
 - Ctrl+Alt+O
 = imports are order in a different order comparing to Google rules - IDEA uses some default settings
 - ./gradlew goJF
 = ordering of imports is fixed in java code as expected

So, I'm too lazy to go to Code Style and fix ordering manually. And I don't want to ask my colleagues to do this as well.

jbduncan:
> Sorry if I started to ask questions in a wrong thread.

No worries! Sorry if I came across poorly.

I think as a general rule, it's best to start a new issue even if an existing closed issue/PR seems relevant because, the way GitHub works, people are more likely to miss conversations on closed issues/PRs.

> In general it's very convenient and useful hot key. You remove unused imports and sort them.
Isn't it a common use case?

I agree! I'm personally used to running the relevant command using Gradle or IntelliJ's support for Gradle, though (as I'm a user/contributor of [Spotless](https://github.com/diffplug/spotless), an alternative plugin which also supports google-java-format).

But having read your comment above, it makes a lot of sense to me that many people prefer using their IDE's built-in command for this. :)

> Do people use such configuration in a real life?

I'm sure people do, but by my understanding it's not recommended for various reasons, i.e. it's better to format the code manually with `./gradlew goJF` before committing, and at CI time to check that all Java source files are formatted properly and fail otherwise.

I wrote about this before on one or two other issues/PRs somewhere on GitHub, but I can't seem to find them. I'll get back to you if I can find them again. :)

PR #96, opened 2016-11-21
Fix imports when formatting by default.

sormuras:
This PR is a proof-of-concept showing the changes needed when the published `format(String)` API fixes import statements by default - instead of leaving them alone.

Except for one not-so-easy to solve issue (see added `Assume.assumeFalse("Skipping " + name + ". Needs fix!", name.equals("Unformatted"));` line in `FormatterIntegrationTest.java`), the changes to the expected outputs are just a few.

See #92 for details and motivation of this PR.

If this route (using `format(String)`) is taken, `FormatFileCallable.call()` needs similar adoption to the new behaviour.


cushon:
Is it OK to drop this in favour of #92, or is there stuff here you'd still like to see addressed?

sormuras:
Totally fine with me. It was a proof-of-concept thing anyway.

Issue #105, opened 2016-12-12
Validation mode to check if file is correctly formatted


JLLeitschuh:
This would be useful in CI systems where you would want the build to fail if the user hasn't formatted their code before the CI system runs.

jbduncan:
FYI [Spotless](https://github.com/diffplug/spotless) (which supports google-java-format) already allows the user to do this. :)

JLLeitschuh:
We are using maven in our build at the moment (trying to move to gradle but transition is slow).
Anything we can do in our world?
I was thinking about doing a checksum diff as a simple fix unless there is a better solution out there.

Stephan202:
We have a `format.sh` script that wraps the Maven plugin, and a [`Dangerfile`](https://github.com/danger/danger) containing the following (this is executed by Travis CI, only for PR builds):

```rb
# Run the formatter and check whether this introduces any changes. 
if !system('./format.sh && git diff --exit-code --name-status')
  fail 'There are formatting errors, please run `./format.sh`.' 
end
```

jbduncan:
@JLLeitschuh Interesting you mention Maven, as we (@nedtwigg and I) are in the middle of writing version 3.0 of Spotless, and one of the things @nedtwigg has been working on recently is splitting out the "core" of Spotless into its own lib so that it can eventually be used in non-Gradle contexts.

However, I don't know if he has any definite plans to support a Maven plugin, so if you're interested, I'd suggest you ask him on the issue he's made to track the spotless-lib efforts. (https://github.com/diffplug/spotless/issues/56)

JLLeitschuh:
I honestly don't plan on supporting our maven build for that much longer either.
I'm in the process of rewriting our build using gradle groovy and [gradle script kotlin](https://github.com/gradle/gradle-script-kotlin) however being a startup with many other things going on simeltaneously I've got a lot of moving targets.

You've been a lot of help.
Do you mind sharing the contents of your `./format.sh` @Stephan202 ? Is it just the call out to the google-java-format jar??

Stephan202:
@JLLeitschuh: there's some other (actually unused) stuff in there, but the no-arg invocation boils down to:
```sh
mvn -B -T 1.0C com.github.PicnicSupermarket:googleformatter-maven-plugin:format
```

[...]

PR #106, opened 2016-12-14
Add dry-run mode and --set-exit-if-changed option

sormuras:
See #105 

`--dry-run` or `-n`
Prints the paths of the files whose contents would change if the formatter were run normally.

`--set-exit-if-changed`
Return exit code 1 if there are any formatting changes.


core/src/main/java/com/google/googlejavaformat/java/CommandLineOptions.java
```diff
@@ -208,7 +221,9 @@ CommandLineOptions build() {
           this.help,
           this.stdin,
           this.fixImportsOnly,
-          this.removeJavadocOnlyImports);
+          this.removeJavadocOnlyImports,
+          this.validationMode
+      );
```
jbduncan: I think this bracket should be on the previous line.

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -251,4 +251,28 @@ public void importRemoveErrorParseError() throws Exception {
...
+  @Test
+  public void validate() throws Exception {
+    Main main =
+        new Main(
+            new PrintWriter(System.out, true),
+            new PrintWriter(System.err, true),
+            System.in);
+    int errorCodeOk = main.format("--validate", //
+        tempJava("Simple", "enum Simple {}\n")//
+    );
```
jbduncan: ditto

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -251,4 +251,28 @@ public void importRemoveErrorParseError() throws Exception {
...
+    assertThat(errorCodeOk).named("Error Code").isEqualTo(0);
+    int errorCode = main.format("--validate", //
+        tempJava("IsFormatted", "interface IsFormatted {}\n"), //
+        tempJava("NeedsFormatting", "  class  NeedsFormatting  {  }\n") //
+    );
```
jbduncan: ditto

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -251,4 +251,28 @@ public void importRemoveErrorParseError() throws Exception {
...
+    assertThat(errorCode).named("Error Code").isEqualTo(1);
+  }
+
+  private String tempJava(String name, String code) throws Exception  {
+    Path path = testFolder.newFile( name + ".java").toPath();
```
jbduncan: There's an unneeded space between the bracket and `name`.

May I suggest running `google-java-format` on the files touched by this PR to make sure everything's nice and compliant? :)

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -251,4 +251,28 @@ public void importRemoveErrorParseError() throws Exception {
...
   }
+
+  @Test
+  public void validate() throws Exception {
+    Main main =
+        new Main(
+            new PrintWriter(System.out, true),
+            new PrintWriter(System.err, true),
+            System.in);
+    int errorCodeOk = main.format("--validate", //
```
jbduncan: Oh, it's also not clear to me why you included blank comments at the end of this line and a few of the lines below. Could you explain why they're there? Are they even needed?

sormuras: No, not needed at all. And gone.

jbduncan: Great, current version looks good to me now. :)

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -151,6 +151,13 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
         allOk = false;
         continue;
       }
+      if (parameters.isValidationMode()) {
+        if (formatted.equals(inputs.get(result.getKey()))) {
+          continue; // no formatting needed, validate next
+        }
+        errWriter.print(result.getKey() + ": needs formatting");
+        return 1; // found first file that needs formatting - exit with error code 1
```
hanwen: can this print all files that need formatting? It's much more useful to get a list of all files that need formatting.

sormuras: Aye. See comment above, at https://github.com/google/google-java-format/pull/106#pullrequestreview-12978555

sormuras:
@ronshapiro @cushon Any chance for the validation mode to get integrated?

f2prateek:
This will be great to have! One thing I'll point out is that in practice I've found it extremely helpful to explain what is expected and what it received — here's how I implemented that in the Gradle Android plugin https://github.com/f2prateek/gradle-android-javafmt-plugin/pull/6#issue-211228769

sormuras:
Seeing a diff is helpful in general -- but when it comes to `google-java-format` you just let it do its work. After a normal run, a failing validation check points to a bug in the formatter. See [FormatterIntegrationTests#idempotent...](https://github.com/google/google-java-format/blob/master/core/src/test/java/com/google/googlejavaformat/java/FormatterIntegrationTest.java) methods.

sormuras:
@cushon One more ping from my side regarding the "validation" feature. Do you think it's worth to have it? I can live with a simple _no_ ... and start using my fork of the project. :)

f2prateek:
That's fair, I do point out in the failure about running fmt, but I also wanted to make it easy for contributors to see what exactly was different.

sormuras:
Rebased.

sormuras:
Rebased.

sormuras:
Rebased.

sormuras:
Rebased.

@cushon @plumpy Any news on having a validation / dry-run mode in the CLI of GJF?

sormuras:
Rebased.

Using https://jitpack.io/com/github/sormuras/google-java-format/validate-SNAPSHOT/ from my fork is an option -- but any answer, be it yes or no, would release me from rebasing that branch. 😉

sormuras:
Rebased to 1.4

sormuras:
- [ ] Yes.
- [ ] No.
- [ ] Next year.

😉

cushon:
Sorry for neglecting this. I asked a clarifying question in #105.

sormuras:
Rebased to 1.5-SNAPSHOT and really like the activity here and in #105 😎

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -142,28 +143,45 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
...
         }
         allOk = false;
         continue;
       }
+      boolean idempotent = formatted.equals(inputs.get(path)); // not changed
+      if (parameters.isValidationMode()) {
+        if (idempotent) {
+          continue; // no formatting needed, validate next
+        }
+        errWriter.println(path + ": needs formatting");
```
cushon: I'd expect stdout for this, and dropping `: needs formatting` would make this easier to use in scripts.

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -142,28 +143,45 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
...
           }
         } else {
-          errWriter.println(result.getKey() + ": error: " + e.getCause().getMessage());
+          errWriter.println(path + ": error: " + e.getCause().getMessage());
           e.getCause().printStackTrace(errWriter);
         }
         allOk = false;
         continue;
       }
+      boolean idempotent = formatted.equals(inputs.get(path)); // not changed
```
cushon: perhaps more self-documenting:

```java
boolean changed = !formatted.equals(inputs.get(path));
```

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -142,28 +143,45 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
...
-          errWriter.println(result.getKey() + ": could not write file: " + e.getMessage());
+          errWriter.println(path + ": could not write file: " + e.getMessage());
           allOk = false;
           continue;
         }
-      } else {
-        outWriter.write(formatted);
+        if (parameters.isSetExitIfChanged()) {
+          allOk = false;
+        }
```
cushon: Can you move this `if` to immediately below the declaration of idempotent/changed to avoid some duplication?

core/src/main/java/com/google/googlejavaformat/java/UsageException.java
```diff
@@ -52,6 +52,10 @@
     "    Character offset to format (0-based; default is all).",
     "  --help, -help, -h",
     "    Print this usage statement",
+    "  --validate, -validate",
+    "    Validate files format only. If files need format print its path to stdout.",
+    "  --set-exit-if-changed, -set-exit-if-changed",
+    "    Validate files only, do not write anything. Error code 0 means everything's formatted.",
```
cushon: I prefer:

```
--dry-run -n
  Prints the paths of the files whose contents would change if the formatter were run normally.
--set-exit-if-changed
  Return exit code 1 if there are any formatting changes
```

validation isn't necessarily the only use of these flags.

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -142,28 +143,45 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
...
+        continue;
+      }
+      outWriter.write(formatted);
+      if (parameters.isSetExitIfChanged() && !idempotent) {
+        allOk = false;
       }
     }
     return allOk ? 0 : 1;
```
cushon: I expected to see changes to formatStdin too?

core/src/main/java/com/google/googlejavaformat/java/UsageException.java
sormuras: Ok.

sormuras: Support `-e` as a short-cut for `--set-exit-if-changed`?

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Ok.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Ok.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Not so easy. The default "stdout" is implemented as a fall-through or "catch-all-else". Will try.

cushon: Sorry, what am I missing?

```
index 7b7c873..b331a78 100644
--- a/core/src/main/java/com/google/googlejavaformat/java/Main.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Main.java
@@ -153,14 +153,14 @@ public final class Main {
         continue;
       }
       boolean idempotent = formatted.equals(inputs.get(path)); // not changed
+      if (!idempotent && parameters.isSetExitIfChanged()) {
+        allOk = false;
+      }
       if (parameters.isValidationMode()) {
         if (idempotent) {
           continue; // no formatting needed, validate next
         }
         errWriter.println(path + ": needs formatting");
-        if (parameters.isSetExitIfChanged()) {
-          allOk = false;
-        }
         continue;
       }
       if (parameters.inPlace()) {
@@ -174,15 +174,9 @@ public final class Main {
           allOk = false;
           continue;
         }
-        if (parameters.isSetExitIfChanged()) {
-          allOk = false;
-        }
         continue;
       }
       outWriter.write(formatted);
-      if (parameters.isSetExitIfChanged() && !idempotent) {
-        allOk = false;
-      }
     }
     return allOk ? 0 : 1;
   }
```

core/src/main/java/com/google/googlejavaformat/java/UsageException.java
cushon: I think it's fine to skip the short form for this one.

sormuras: Ok.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: I'll push a cleaner solution. It needs logical "is-not-running-in-special-mode" helper in `CommandLineOptions`:

```java
  /** Returns true if no special run-mode (like in-place or dry-run) is selected. */
  boolean stdout() {
    return !(inPlace || dryRun);
  }
```

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Not sure about this one.

Should an exit code of 1 be returned if changes where made for input from stdin and `--set-exit-if-changed` is set to true?

core/src/main/java/com/google/googlejavaformat/java/Main.java
cushon: Making the combination a usage error sounds fine.

I'm still not sure I understand what the help is needed for?

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -142,28 +143,36 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
...
         }
         allOk = false;
         continue;
       }
+      if (parameters.stdout()) {
+        outWriter.write(formatted);
+      }
+      boolean changed = !formatted.equals(inputs.get(path));
+      if (!changed) {
+        continue; // preserve original file
```
sormuras: Could be transformed to

```java
if (changed) {
...
}
```

for clarity. Renders the `continue` obsolete.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Maybe I'm assuming s/t wrong here. 

Should the stdout-case **always** print out the potentially formatted text?

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: You are referring to the method `private int formatStdin(CommandLineOptions parameters, JavaFormatterOptions options) {...}`!

core/src/main/java/com/google/googlejavaformat/java/Main.java
cushon: `google-java-format T.java` always prints formatted text to stdout, regardless of whether the input was formatted. `google-java-format -n T.java` prints `T.java` to stdout if the input was not already formatted.

core/src/main/java/com/google/googlejavaformat/java/Main.java
cushon: I would expect `-n` and `--set-exit-if-changed` to work the same for input from stdin as it does for files. For `google-java-format -n -` we can reuse `STDIN_FILENAME`.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: The current iteration behaves like this. I don't see an easy way to re-use a boolean state at the moment.

core/src/main/java/com/google/googlejavaformat/java/Main.java
sormuras: Ok. On it.

sormuras: Done via:
```java
      if (parameters.isDryRun()) {
        outWriter.println(STDIN_FILENAME);
      } else {
        outWriter.write(output);
      }
      if (input.equals(output)) {
        return 0;
      }
      return parameters.isSetExitIfChanged() ? 1 : 0;
```

sormuras:
I'll leave the PR now as-is. If anything needs a tweak or correction, let me know.

Cheers!

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -178,8 +187,15 @@ private int formatStdin(CommandLineOptions parameters, JavaFormatterOptions opti
     }
     try {
       String output = new FormatFileCallable(parameters, input, options).call();
-      outWriter.write(output);
-      return 0;
+      if (parameters.isDryRun()) {
+        outWriter.println(STDIN_FILENAME);
```
cushon: Let's only do this if the input was unformatted, same as for file inputs.

sormuras: Ok.

sormuras:
Cleaned up and tests added.

cushon:
Merged as 13354f3ca1081ad0618f75e6826ac96c19f404be.

I made some adjustments but nothing that should have changed behaviour. Let me know if I missed anything, and thanks for your work on this!

sormuras:
Build via https://jitpack.io/com/github/google/google-java-format/master-SNAPSHOT/build.log
Download `all-deps` from https://jitpack.io/com/github/google/google-java-format/google-java-format/master-SNAPSHOT/google-java-format-master-SNAPSHOT-all-deps.jar

`java -jar google-java-format-master-SNAPSHOT-all-deps.jar --version`

yields:

`google-java-format: Version 1.5-SNAPSHOT`

And running `java -jar google-java-format-master-SNAPSHOT-all-deps.jar --dry-run --set-exit-if-changed` on a tree of un-formatted sources exits with error code `1` as expected.

✔️

sormuras:
Updating my project to consume GJF 1.5-SNAPSHOT instead of my fork was trivial: https://github.com/sormuras/bach/commit/99d48cb5a3c0dfb18eb5b24b83e95e09127f1498

Issue #119, opened 2017-02-09
Eclipse plugin produces different result as standalone formatter version


davido:
I'm testing the plugin, by installing it and selecting the `google-java-formatter`: [1].
Trying it on [this](https://gerrit.googlesource.com/gerrit/+/master/gerrit-acceptance-tests/src/test/java/com/google/gerrit/acceptance/edit/ChangeEditIT.java) file in Gerrit Code Review, produces completely different result, as release
version 1.2 of google java-formatter: [2]. What I was doing, is `Ctrl-A` followed by `Ctrl+Shift+F`.

If I run this command:

```
$ google-java-format -i gerrit-acceptance-tests/src/test/java/com/google/gerrit/acceptance/edit/ChangeEditIT.java
```

where the reported version is:

```
$ google-java-format --version
google-java-format: Version 1.2
```

then I get the commited version again (no diff).

Eclipse version:

```
Version: Neon.1a Release (4.6.1)
Build id: 20161007-1200
```

google-java-formatter is built from: 760a13c9cc7c5741c289dbea8162970e9b33bdbe

* [1] http://imgur.com/a/TXliV
* [2] http://paste.openstack.org/show/598171/

sormuras:
Just guessing, the max line length is set to 80 and somehow that setting is applied to the formatted source.

sormuras:
And/or _Gerrit Code Review_ does not honor the custom formatter implementation.

What happens, if you open `ChangeEditIT.java` in an editor and press just `Ctrl+Shift+F`?

cushon:
That output wasn't produced by google-java-format. It looks like the built-in eclipse formatter.

The plugin worked for me with 4.6.2.

davido:
> And/or Gerrit Code Review does not honor the custom formatter implementation.

What do you mean exactly here? It's just a workspace that I opened in Eclipse and replaced the formatter.

Can it be some issues in my workspace then? Of course the next attempt is to upgrade my Eclipse to 4.6.2.

davido:
Upgrade to Eclipse didn't help either. Ctrl-Shift-F, with Ctrl-A doesn't format with gjf.

I have an impression, that gjf only works in the Preferences->Java-Code Style=>Formatter, but not in the editor.

[...]

PR #120, opened 2017-02-09
Prevent edit if nothing was formatted

sormuras:
Handle trivial cases, where input equals output, to prevent dirty flags in active editors.

### DIFF
--- a/eclipse_plugin/src/com/google/googlejavaformat/java/GoogleJavaFormatter.java
+++ b/eclipse_plugin/src/com/google/googlejavaformat/java/GoogleJavaFormatter.java
@@ -86,22 +86,24 @@ public class GoogleJavaFormatter extends CodeFormatter {
           new SnippetFormatter()
               .format(
                   snippetKind, source, rangesFromRegions(regions), initialIndent, includeComments);
-      // Trivial case: no replacement, no change.
+      // Trivial case: no replacement, nothing changed.
       if (replacements.isEmpty()) {
         return null;
       }
-      // Entire source case: input = output, no change.
-      if (replacements.size() == 1 && replacements.get(0).getReplacementString().equals(source)) {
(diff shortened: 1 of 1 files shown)


cushon:
This looks ok. Can you elaborate on the problem it solves? Is it related to #119?

sormuras:
It was discovered while reproducing the problem described in #119 

When you format an entire file (or section) once and save it, a subsequent format should not change the text. Idempotency like ensured by https://github.com/google/google-java-format/blob/master/core/src/test/java/com/google/googlejavaformat/java/FormatterIntegrationTest.java#L101

The current implementation replaces the old text with new text, even when they are equal, and that marks the active editor as dirty.

sormuras:
#105 and #106 are related. In a console-y way. 😉

sormuras:
Just give me another second, moving the new code into its own method.

davido:
Oh, I see this problem, now that the plugi was fixed to work against Gerrit project. After pressing Ctrl-Shift+F without changes, the file is markes durty, and asterisk appear in fron of the file name.

davido:
@sormuras 

With this CL applied, the pugis works as expected. There is no differentce between `Ctrl+Shift+F` and gjf invocation from the command line. That's really really great. Thank you so much.

@cushon 
Can we release new gjf version including the plugin binary ASAP?

cushon:
Sure, what's the best way to release this? I don't know how eclipse plugins are distributed. Would being able to download it from the github release page be a useful place to start?

davido:
> Would being able to download it from the github release page be a useful place to start?

That would be cool! I volunteer to upload change to Gerrit with instructions, how to deploy it and write announcement on the repo-discuss mailing list. Thank you again for your help!

sormuras:
You're welcome!

Issue #125, opened 2017-02-11
Use eclipse style sorting order for static imports


davido:
Even though, the ordinal number of capitalized letters is higher than normal letter, the soring order for natural text is usually differs. This is also te default behaviour of sort command in *nices:
   
``` 
      $ cat file
      A
      a
      $ cat file | sort
      a
      A
```

The main reason for changing the sorting behaviour here, to match that for natural language text, is to get exactly the same behaviour as in Eclipse's IDE "Organize Imports" feature.

For more context, see the discussion in: [1], [2].

* [1] https://gerrit-review.googlesource.com/96950
* [2] https://gerrit-review.googlesource.com/97054

sormuras:
Does the natural ordering adhere to https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing ?

cushon:
The current behaviour is required by the style guide: https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing. There's been some discussion about the merits of Eclipse's approach, but there are currently no plans to change the style guide.

The other difference I'm aware of is that Eclipse groups static imports by class instead of sorting the entire qualified name:

```java
import static a.Foo.FOO;
import static a.Foo.ZOO;
import static a.Foo.SubFoo.BAR;
```

vs.

```java
import static a.Foo.FOO;
import static a.Foo.SubFoo.BAR;
import static a.Foo.ZOO;
```

davido:
> The current behaviour is required by the style guide: https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing. There's been some discussion about the merits of Eclipse's approach, but there are currently no plans to change the style guide.

Quoting the spec here:

```
[...]
Within each block the imported names appear in ASCII sort order.
```

Would it be an option to deviate from the spec guarded by an extra option, say `gjf -i -natural-sort-ordering-deviate-from-thespec-i-know-what-i-am-doing Foo.java`, as non working `Ctrl-Shift-O` in Eclipse horribly sucks.

cushon:
The [examples](http://paste.openstack.org/show/598484/) in your second context link are both caused by the different handling of static imports. As far as I can tell Eclipse also uses ASCII order.

How often are you seeing this happen? Internally it affects about one file in six thousand, and we have a pre-commit check to make sure the order is correct. It hasn't been a significant issue in practice.

davido:
[...]

PR #126, opened 2017-02-11
Use container first sorting order for static imports 

davido:
Use container first sorting order for static imports 

Fixes: #125


davido:
I signed it!

cushon:
As discussed in #125 we'd prefer eclipse support google style rather than changing the formatter or adding configuration.

PR #131, opened 2017-02-24
Remove redundant `String::format` call.

jbduncan:


jbduncan:
I do realise that this PR came out of the blue without any previous discussion, but I'm hoping that it's reasonably obvious as to what I'm hoping to achieve with it, and so I'd like to begin discussing with someone on the google-java-format team to see if it has any merit or if it needs changing/rejecting. :)

cushon:
This was fixed in 13354f3ca1081ad0618f75e6826ac96c19f404be.

jbduncan:
> This was fixed in 13354f3.

Oh, cool! Thanks for letting me know @cushon.

Thanks for fixing it up for me @sormuras. :)

PR #133, opened 2017-02-24
Swap out JDK singleton list usages with ImmutableList

jbduncan:
This helps to make the code a little more consistent and maintainable, since all singleton lists now use the same class.


jbduncan:
Many thanks for merging this in @cushon. :+1:

PR #138, opened 2017-03-16
Expand IntelliJ section in README to suggest formatter style.xml

meonlol:
Because it is not really practical to tell our team members to manually format with the google-java-formatter one time after every start-up, I looked for a different solution and finally found the intellij style.xml.

This pull-request is for expanding the README to suggest this for others who are also faced with the same situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ files. Until IntelliJ is restarted, the `Reformat code` action will also
 There is an [open bug](https://devnet.jetbrains.com/thread/464297) against
 IntelliJ to add support for configuring external formatters.
 
+If this is currently not an option for your team, you could also use the
+formatter 
+[rule set](https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml)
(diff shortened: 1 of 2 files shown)


kevinb9n:
I am not sure this is good advice. We went to all the trouble to build this product because IntelliJ's code formatting is not that great.

meonlol:
(Closed it because I wanted to make additional changes. (Got fierce pushback on the 100 column width))

I understand that and I agree, but the guys at IntelliJ do not seem to be motivated to change it, and in the current form it's not really practical.
Or do you have another way to persist the settings over a restart? Or do you perhaps use macro-like save actions?

talios:
I provided a PR some time ago with exposed the internal line-length setting ( which used to work in 1.0, then broke in 1.1 ) which I used in my https://github.com/talios/googleformatter-maven-plugin project, but there was little interest in the patch, even tho the 1.1 release broke the behaviour ).

Nice to see it being increased to 150 tho ( we tend to use 160 at work ).

PR #145, opened 2017-04-25
Make Classes Main and CommandLineOptions usable from within own code

jpstotz:
A lot of useful code could not be used because the classes and their methods has not been marked as public. I don't see a reason why one should re-implement all of this. Therefore I added the public modifier to certain classes, some builder classes and the builder methods as well as some methods in the `Main` class.
Those classes and methods are necessary when using google-java-format from within own code.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptions.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptions.java
@@ -23,7 +23,7 @@ import com.google.common.collect.ImmutableRangeSet;
  * <p>google-java-format doesn't depend on AutoValue, to allow AutoValue to depend on
  * google-java-format.
  */
-final class CommandLineOptions {
+public final class CommandLineOptions {
 
   private final ImmutableList<String> files;
   private final boolean inPlace;
@@ -145,7 +145,7 @@ final class CommandLineOptions {
     return new Builder();
   }
 
-  static class Builder {
+  public static class Builder {
 
     private final ImmutableList.Builder<String> files = ImmutableList.builder();
     private final ImmutableRangeSet.Builder<Integer> lines = ImmutableRangeSet.builder();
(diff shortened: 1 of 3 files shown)


jpstotz:
I have signed the CLA

core/src/main/java/com/google/googlejavaformat/java/CommandLineOptions.java
```diff
@@ -145,7 +145,7 @@ static Builder builder() {
     return new Builder();
   }
 
-  static class Builder {
+  public static class Builder {
```
NightlyNexus: `final`.

jbduncan:
@jpstotz Can you explain to me why `CommandLineOptions` and `Main.formatFiles(CommandLineOptions)` are needed? It's not clear to me from your description, so would you be at liberty to explain further?

jpstotz:
When using google-java-format as a library there are two options for using a built `CommandLineOptions`:

1. Use it in a `FormatFileCallable` 
2. Use it with `Main.formatFiles(CommandLineOptions)`.

First I started to use the `FormatFileCallable` together with an `ExecutorService`. Then I realized that `Main.formatFiles(CommandLineOptions)` has already everything in place for formatting a bunch of files. 
As I prefer to use well-engineered existing code rather than re-inventing the wheel I made the method public...

jbduncan:
Okay, thanks for clearing things up for me @jpstotz.

Have you considered exposing and moving the functionality of `Main.formatFiles(CommandLineOptions)` in a method in a separate, more descriptive class?

Have you also considered removing the direct dependency on `CommandLineOptions` and replacing it with a `String[]` or `List<String>` of command line options instead?

jpstotz:
@jbduncan My assumption was that small changes have a higher probability of acceptance - therefore I minimized the changes. Regarding your suggestion the command-line options via `List<String>`: I am sorry but that something I excluded right for the beginning as using command-line options within a program is just hacky. IMHO when a Java program is using a Java library they should communicate via data structures, not  command-line options designed for humans. Using a Java library in a Java program via String based command-line options is similar as two native English speakers that talk to each other using Latin...

@cushon If I understand the usage of `com.google.googlejavaformat.java.Formatter` correctly this is an entry point which has been designed for the usage of google-java-format as a library within a plugin of an Java development IDE. For batch processing of files you need to re-implement a lot of code that is already present in other parts the library.

jbduncan:
@jpstotz Did you mean to sound loaded just now? :)

I agree with your argument, especially with regards to using the right input data structure! For that reason, I'd argue further that `CommandLineOptions` is the wrong structure to use (assuming that I've understood what you're asking for), since an instance of that class contains a lot of information which isn't needed in the context of formatting many files across multiple cores.

FWIW, I wrote my last response from my phone, so I wasn't able to check that any of my suggestions would have made sense code-wise, so apologies if I came across poorly.

Now that I've had time to look over `google-java-format` properly again on my dev machine, I wonder if I may suggest an alternative.

From what I've gathered, you're after the ability to be able to format multiple files across multiple cores. Have I understood this correctly?

If so, I would suggest to look into extracting the logic from `Main.formatFiles()` (which was designed to be an internal method AFAICT) that deals with formatting multiple files concurrently on an `ExecutorService`, and then put that logic into a new method in `Formatter` like the following code skeleton.

```java
public final class Formatter {
  . . .

  public List<String> formatSources(List<String> inputs, ExecutorService executor) {
    . . .
  }
}
```

...I've just copied the contents of `Main.formatFiles()` into this method skeleton on my local copy of google-java-format, and I've realised that unfortunately it's not quite as simple filling out this skeleton since, to give a couple of examples, `Main.formatFiles()` itself depends on an `errWriter` to log errors, and also one of its dependencies `FormatFileCallable` expects a `CommandLineOptions` instance at construction time. So there would have to be quite a lot of refactoring involved...

I hope this helps.

(@cushon, what do you think of this suggestion?)

jbduncan:
@jpstotz To add to my last message: yes, there is a higher chance of a change being accepted if it's small, but I think @cushon would agree with me in saying that your current proposed change is a bit dangerous, because it exposes more implementation details to the users of the library side of `google-java-format`, which then makes it harder for Google to change those implementation details if or when the time comes.

sormuras:
Related to #561 which would offer the same functionality by using the standardized [ToolProvider](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/spi/ToolProvider.html) SPI.

PR #173, opened 2017-08-11
Upgrade to JUnit 5 and Surefire 2.19.1

sormuras:
This PR introduces **JUnit 5** as a dependency and ports a single test class, namely `NewlinesTest.java`, to the new [JUnit Jupiter API](http://junit.org/junit5/docs/current/user-guide/).

### Todo
- [ ] Convert all remaining test classes to JUnit 5

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -59,6 +59,12 @@
 
     <!-- Test dependencies -->
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <version>5.0.0-RC2</version>
+      <scope>test</scope>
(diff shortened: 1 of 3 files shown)


cushon:
This will have to wait until we start using JUnit 5 internally, which could take a while.

sormuras:
Alright. Will keep the version(s) up-to-date and convert the rest of the tests when you're readsy to switch.

cushon:
Unfortunately I don't have an ETA, we're going to be using JUnit 4 for the foreseeable future. I'm going to close this rather than leave it open indefinitely, but I'm happy to revisit if/when we make the switch.

sormuras:
No problem.

PR #177, opened 2017-08-20
Add (ignored) test cases for Java 9 syntax related issues

sormuras:
This PR adds file `Java9SourceTest.java` to show the problems with Java 9 syntax features and the new `javac` version included in https://github.com/sormuras/google-java-format/commit/fee1ed2685f0b4b60cafde2b5204b11ad5bc69c5

- #75 
- #155 
- #176 

This PR also fixes a typo in `TryWithResources.input` and `TryWithResources.output`.

PR #178, opened 2017-08-20
Set supported Java source level to "9"

### LINKED ISSUE #176: Java 9 - private interface methods are not supported
Since the update of the "javac" dependency in https://github.com/google/google-java-format/commit/fee1ed2685f0b4b60cafde2b5204b11ad5bc69c5, formatting an interface that defines private static methods fails with the following error message:

```
 Formatter crashed on InterfaceWithPrivateStaticMethod: 2:27: error: private interface methods are not supported in -source 1.8
  (use -source 9 or higher to enable private interface methods)
```

Interface source:
```java
interface InterfaceWithPrivateStaticMethod {
  private static void foo() {}
}
```

Is there a simple way pass the "-source 9" to the formatter instance (via cli)?

sormuras:
Fixes #176 by setting the default javac source level to "9".

This PR superseds #177 as it includes the failing and ignored as well.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Formatter.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Formatter.java
@@ -47,6 +47,7 @@ import org.openjdk.javax.tools.JavaFileObject;
 import org.openjdk.javax.tools.SimpleJavaFileObject;
 import org.openjdk.javax.tools.StandardLocation;
 import org.openjdk.tools.javac.file.JavacFileManager;
+import org.openjdk.tools.javac.main.Option;
 import org.openjdk.tools.javac.parser.JavacParser;
 import org.openjdk.tools.javac.parser.ParserFactory;
 import org.openjdk.tools.javac.tree.JCTree.JCCompilationUnit;
@@ -135,6 +136,10 @@ public final class Formatter {
     DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
     context.put(DiagnosticListener.class, diagnostics);
     Options.instance(context).put("allowStringFolding", "false");
+    // TODO(cushon): this should default to the latest supported source level, remove this after
+    // backing out
+    // https://github.com/google/error-prone-javac/commit/c97f34ddd2308302587ce2de6d0c984836ea5b9f
+    Options.instance(context).put(Option.SOURCE, "9");
     JCCompilationUnit unit;
     JavacFileManager fileManager = new JavacFileManager(context, true, UTF_8);
     try {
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -135,6 +135,7 @@ static void format(
     DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
     context.put(DiagnosticListener.class, diagnostics);
     Options.instance(context).put("allowStringFolding", "false");
+    Options.instance(context).put("-source", options.javaSourceLevel());
```
cushon: Please add a TODO so I remember to remove this once the default language level is 9. Also you can use the `Option` enum instead of the string `"-source"`:

```java
// TODO(cushon): this should default to the latest supported source level, remove this after backing out
// https://github.com/google/error-prone-javac/commit/c97f34ddd2308302587ce2de6d0c984836ea5b9f
Options.instance(context).put(Option.SOURCE, "9");
```

core/src/main/java/com/google/googlejavaformat/java/JavaFormatterOptions.java
```diff
@@ -30,6 +30,7 @@
 public class JavaFormatterOptions {
 
   static final int DEFAULT_MAX_LINE_LENGTH = 100;
+  static final String DEFAULT_JAVA_SOURCE_LEVEL = "9";
```
cushon: This doesn't need configuration. It would normally default to the latest supported language level, which is want we want. Explicitly setting `-source 9` is a temporary work-around (see the TODO in my other comment).

core/src/test/java/com/google/googlejavaformat/java/Java9SourceLevelSupportTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import com.google.common.base.Joiner;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for Java 9 source level related issues. */
+@RunWith(JUnit4.class)
+public class Java9SourceLevelSupportTest {
```
cushon: Let's hold off on these tests until these features are implemented. I'd rather use the regular side car files for the test inputs.

sormuras: Alright. I'll remove the file from the PR.

Btw, is somebody already coding the `visitModule(ModuleTree node...` method and friends? If not, I could give it shot.

sormuras:
Requested changes applied.

sormuras:
Travis CI is having troubles starting the PR build...

PR #195, opened 2017-09-23
fix locale problem on DiagnosticTest

winteryoung:
fixed a test bug that caused under none english locale;
fixed a potential bug;
removed an unused parameter;


core/src/test/java/com/google/googlejavaformat/java/DiagnosticTest.java
```diff
@@ -37,6 +40,11 @@
 public class DiagnosticTest {
   @Rule public TemporaryFolder testFolder = new TemporaryFolder();
 
+  @BeforeClass
+  public static void setUpLocale() throws Exception {
+    Locale.setDefault(Locale.US);
+  }
```
jbduncan: I think it would be much better to pass `Locale.US` to the methods that require them instead of making it the global default. That way, other tests which expect a different default locale would be less likely to fail.

core/src/main/java/com/google/googlejavaformat/InputOutput.java
```diff
@@ -82,7 +82,7 @@ protected final void computeRanges(List<? extends Input.Tok> toks) {
         for (int i = lineI0; i <= lineI; i++) {
           addToRanges(ranges, i, k);
         }
-        addToRanges(range1s, lineI0, k);
+        addToRanges(range1s, lineI, k);
```
jbduncan: @winteryoung Could you explain for me why you changed the argument from `lineIO` to `lineI` here? It's not immediately obvious to me why you think it's correct.

winteryoung: The original code is
```java
      if (k >= 0) {
        addToRanges(range0s, lineI0, k);
        for (int i = lineI0; i <= lineI; i++) {
          addToRanges(ranges, i, k);
        }
        addToRanges(range1s, lineI0, k);
      }
```
The first `addToRanges()` and the last `addToRanges()` make no difference. The original one could pass the tests. I guess that's because there's no test cases contain tokens that span multiple lines.

cushon: This doesn't look right to me. I think this doesn't break tests because `range0s` isn't being used, and `range1s` is being used in a situation where `ranges` would also be correct.

core/src/test/java/com/google/googlejavaformat/java/DiagnosticTest.java
cushon: The USian part is that the `google-java-format` diagnostics contain javac diagnostics for parse errors, and javac diagnostics are localized (e.g. this breaks if you set `Locale.JAPANESE`).

Maybe we should override the locale when formatting the javac diagnostics? I thought that was already happening, but apparently it doesn't work reliably: https://github.com/google/google-java-format/blob/fee1ed2685f0b4b60cafde2b5204b11ad5bc69c5/core/src/main/java/com/google/googlejavaformat/FormattingError.java#L56

core/src/main/java/com/google/googlejavaformat/InputOutput.java
winteryoung: Then what's the difference between `range0s` and `range1s`? They contain the same contents.

cushon: Ah, thanks. This does look like a mistake, but the reason it wasn't breaking tests is that it wasn't really being used. I removed it in c48ee99f0240e0e51af842b3cf17793a83508bd4.

winteryoung:
All done

winteryoung:
I've done all you guys asked. Let's give this issue a closure.

core/src/test/java/com/google/googlejavaformat/java/DiagnosticTest.java
```diff
@@ -26,8 +26,9 @@
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import org.junit.Rule;
-import org.junit.Test;
+import java.util.Locale;
+
+import org.junit.*;
```
jbduncan: I believe wildcard imports are disallowed in the Google Java Style guide. Have you run `google-java-format` on your changes already?

idea_plugin/google-java-format.iml
```diff
@@ -13,6 +13,6 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="library" name="google-java-format-1.4-all-deps" level="project" />
+    <orderEntry type="library" name="google-java-format-1.5-SNAPSHOT-all-deps" level="project" />
```
jbduncan: I admit it's not clear to me why google-java-format was "upgraded" to a SNAPSHOT version. Can you explain your reasoning to me here?

idea_plugin/google-java-format.iml
```diff
@@ -13,6 +13,6 @@
...
   </component>
-</module>
+</module>
```
jbduncan: Minor nitpick: I believe the EOL should stay here. :)

core/src/test/java/com/google/googlejavaformat/java/DiagnosticTest.java
```diff
@@ -37,6 +38,19 @@
...
+  public void setUpLocale() throws Exception {
+    backupLocale = Locale.getDefault();
+    Locale.setDefault(Locale.ROOT);
+  }
+
+  @After
+  public void restoreLocale() throws Exception {
+    Locale.setDefault(backupLocale);
+  }
+
```
jbduncan: Looks much better, many thanks! :+1:

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -238,6 +242,9 @@ public void importRemovalLines() throws Exception {
   // test that errors are reported on the right line when imports are removed
   @Test
   public void importRemoveErrorParseError() throws Exception {
+    Locale backupLocale = Locale.getDefault();
+    Locale.setDefault(Locale.ROOT);
+
```
jbduncan: :+1:

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -253,6 +260,8 @@ public void importRemoveErrorParseError() throws Exception {
             new ByteArrayInputStream(joiner.join(input).getBytes(UTF_8)));
     assertThat(main.format("-")).isEqualTo(1);
     assertThat(err.toString()).contains("<stdin>:4:3: error: class, interface, or enum expected");
+
+    Locale.setDefault(backupLocale);
```
jbduncan: :+1:

core/src/test/java/com/google/googlejavaformat/java/DiagnosticTest.java
```diff
@@ -26,6 +26,9 @@
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Locale;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
```
winteryoung: I used google-java-format command line util on this file specifically against this file this time. It seems the import rules do allow wild card imports. I had to re-import every junit dep manually.

jbduncan: > It seems the import rules do allow wild card imports.

Oh, really? It looks to me that point [3.3.1](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports) of the guide says that "**Wildcard imports**, static or otherwise, **are not used**."

Or have I misunderstood what you mean by "import rules"?

winteryoung: Oh. I mean the rules coded in the source of google-java-format. I'm not familiar with the google java style guide. I am a fan of tools. That's why I came to google-java-format.

jbduncan: Oh, how funny. I just tried it myself, and indeed google-java-format doesn't reformat wildcard imports as non-wildcard imports. I must've got confused with how google-java-format automatically removes unused imports. Apologies!

But regardless, thank you very much for removing the wildcards that were here previously. :)

winteryoung: No problem :)

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -238,6 +242,9 @@ public void importRemovalLines() throws Exception {
   // test that errors are reported on the right line when imports are removed
   @Test
   public void importRemoveErrorParseError() throws Exception {
+    Locale backupLocale = Locale.getDefault();
+    Locale.setDefault(Locale.ROOT);
```
cushon: Do we understand why the diagnostics are being affected by the default locale. As I mentioned earlier, that isn't supposed to happen: https://github.com/google/google-java-format/pull/195/commits/996962604f2b772a14aa9729d9fc8ae7d9dd7df1#r140948765

jbduncan: Hmm... Sorry @cushon, but I don't see a comment when I click on that link, so it's not clear to me what you're referring to. :(

cushon: Do this work?

`https://github.com/google/google-java-format/pull/195/commits/996962604f2b772a14aa9729d9fc8ae7d9dd7df1#r140948765`

jbduncan:
@cushon Yep, that link works for me, thanks!

I don't know enough to know why the locale isn't being consistently overridden to a specific value like `Locale.ENGLISH`, so I fear I'll be off little help here.

cushon:
I think there's still an underlying problem that diagnostics are only getting partially localized - the bits that come from javac are localized, the rest aren't. It would be better to be consistent (and better still to actually localize all of the diagnostics...), but it's fine to merge this as-is to avoid the test failure for now.

PR #199, opened 2017-09-29
idea_plugin README: Add macOS menu location to instructions

apjanke:
This PR updates the `idea_plugin/README.md` instructions to reflect the menu layout on the macOS version of IntelliJ IDEA.

It also changes the checkbox name from "Enable" to "Enable google-java-format", since that's the full checkbox label in the current (2017.2.5) version of IDEA (in both macOS and Windows).


apjanke:
Thanks for the rapid response!

PR #206, opened 2017-10-09
Don't introduce a space for //noinspection

### LINKED ISSUE #202: GJF rewrites IntelliJ / Android Lint noinspection suppressions
Both IntelliJ and Android Lint support doing warning suppressions on a single line with a `//noinspection` comment, e.g.:
```java
//noinspection CheckResult
methodWhoseResultShouldBeChecked();
```
We have many such suppressions in our code base.  GJF rewrites 
`//noinspection` to `// noinspection` and unfortunately Android Lint does not recognize the suppression after the rewrite.  Even with Java 8 there is no built-in suppression mechanism that can be applied to any arbitrary line of code, so I don't see a great alternative here.

The Google Java style guide seems [fairly clear](https://google.github.io/styleguide/javaguide.html#s4.8.6-comments) that there should be a space after `//` comments.  But is it possible an exception could be made here?  If not, are there any suggested workarounds?

msridhar:
This reformatting breaks Android Lint and IntelliJ suppressions

Fixes #202

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/JavaCommentsHelper.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/JavaCommentsHelper.java
@@ -115,10 +115,10 @@ public final class JavaCommentsHelper implements CommentsHelper {
         // If it is `//noinspection` then don't add a space.  `//noinspection` is
         // used for single-line suppressions by both IntelliJ and Android Lint,
         // and for now there is no good alternative.
-        if ((line.length() > length + 12)
-            && line.substring(length,length+12).equals("noinspection")) {
-          // do nothing
-        } else {
+        boolean hasNoSuppression =
+            (line.length() > length + 12)
+                && line.substring(length, length + 12).equals("noinspection");
+        if (!hasNoSuppression) {
           line = Strings.repeat("/", length) + " " + line.substring(length);
         }
       }
@@ -180,4 +180,3 @@ public final class JavaCommentsHelper implements CommentsHelper {
     return true;
   }
 }
-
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/googlejavaformat/java/JavaCommentsHelper.java
```diff
@@ -112,7 +112,15 @@ private String indentLineComments(List<String> lines, int column0) {
       Matcher matcher = LINE_COMMENT_PREFIX.matcher(line);
       if (matcher.find()) {
         int length = matcher.group(1).length();
-        line = Strings.repeat("/", length) + " " + line.substring(length);
+        // If it is `//noinspection` then don't add a space.  `//noinspection` is
+        // used for single-line suppressions by both IntelliJ and Android Lint,
+        // and for now there is no good alternative.
+        boolean hasNoSuppression =
+            (line.length() > length + 12)
+                && line.substring(length, length + 12).equals("noinspection");
```
JakeWharton: `regionMatches`?

core/src/main/java/com/google/googlejavaformat/java/JavaCommentsHelper.java
```diff
@@ -112,7 +112,15 @@ private String indentLineComments(List<String> lines, int column0) {
       Matcher matcher = LINE_COMMENT_PREFIX.matcher(line);
       if (matcher.find()) {
         int length = matcher.group(1).length();
-        line = Strings.repeat("/", length) + " " + line.substring(length);
+        // If it is `//noinspection` then don't add a space.  `//noinspection` is
+        // used for single-line suppressions by both IntelliJ and Android Lint,
+        // and for now there is no good alternative.
+        boolean hasNoSuppression =
```
JakeWharton: This should either be `hasSuppression` or `hasNoInspection`. Currently the naming is inaccurate since it's true when there is a suppression.

msridhar:
Closing since @cushon has [already fixed this internally](https://github.com/google/google-java-format/issues/202#issuecomment-335313873)

PR #208, opened 2017-10-10
Optimize newline processing

novalis:
This makes google-java-format as a whole run in about a third of the
time on my (already-correctly-formatted) codebase.

I could probably use a cache with an LRU eviction policy here, but it's not clear what size is reasonable.  I could also do a per-file cache, but (a) often most files start with the same copyright header, and (b) the most-frequently-used item in the cache is " ", so there's some between-files sharing.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/Newlines.java
+++ b/core/src/main/java/com/google/googlejavaformat/Newlines.java
@@ -16,23 +16,27 @@ package com.google.googlejavaformat;
 
 import com.google.common.base.CharMatcher;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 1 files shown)


cushon:
Thanks for catching this. I think there's an underlying bug that's causing this code to be called more than it should be.

Can you you how the performance of this fix compares to https://github.com/cushon/google-java-format/commit/b9751f39b442b4b3024abea0654c4d1a384e0408 on your codebase?

novalis:
Yours is better (7s vs 8s on my random test codebase).  I'm closing this PR, on the assumption that you'll merge yours.

PR #213, opened 2017-10-14
Reset to 1.5-SNAPSHOT

sormuras:
The cleanup commit https://github.com/google/google-java-format/commit/2ec27b53b185e828db25f89b6e99ca8d520e5fcb set it back to `1.0`...


cushon:
Argh. Thanks for spotting this, it's an issue with our sync process.

I'm working on a change to read the version from `META-INF/maven/com.google.googlejavaformat/google-java-format/pom.properties` at runtime instead of trying to keep it up to date by hand, sort of like [1]. Sound OK?

[1] https://github.com/google/error-prone/blob/6bf224e92d632c8485f424f3c2faefa3b1863b27/check_api/src/main/java/com/google/errorprone/ErrorProneVersion.java#L36

sormuras:
Sounds good. Closing this PR in favor of your dynamic version.

PR #216, opened 2017-10-23
Introduce support for usage with Bazel.

abergmeier-dsfishlabs:


abergmeier-dsfishlabs:
@damienmg CC you for comments.

damienmg:
ball to @iirina :)

dhalperi:
FYI this would be rad, also would probably require a release for us to use it (so that others not using bazel have easy a release jar that has the same behavior).

t-rad679:
Any update on this? This would be a very nice feature and it looks like it's complete. Is there a reason it hasn't been merged yet?

rkhir:
Hello, is there any update on this branch? we really need a feature instead of running the .jar formatter

davido:
See also: https://github.com/bazel-contrib/rules_jvm/issues/23.

Issue #221, opened 2017-11-07
Enforcing a space in one-line comment breaks //$NON-NLS-x$ notation


romp13:
Hi,
Since version 1.4 a new rule formats "//comment" into "// comment". Unfortunately Eclipse uses "//$NON-NLS-x$" comments to avoid internationalization warnings. If you add the space it does not work anymore. We have these all over the place in our codebase which prevents us of using versions of google-java-format later than 1.3.
Would it be possible to handle this case? (maybe a solution is to check if the first character is alpha-numeric)
Alternatively would there be a workaround for us with a setting.
Thanks!

briandealwis:
By way of background, the Eclipse Compiler for Java (ecj) can be configured to warn about non-externalized strings for internationalization.  The warnings can be suppressed by annotating strings with comments like `//$NON-NLS-x$` to indicate that the `x`th string on the line is not meant to be externalized (such as an identifier).

An alternative approach to #222 would be to special case `$NON-NLS-x$` similar to the `noinspection` that's already supported:
```
private static final Pattern LINE_COMMENT_MISSING_SPACE_PREFIX =
      Pattern.compile("^(//+)(?!noinspection|\\$NON-NLS-\\d+\\$)[^\\s/]");
```

However there's one additional difference in that the `//$NON-NLS-x$` should be kept with the string, and it shouldn't wrap even if it extends beyond the line-length.  But perhaps that should just be punted to the developers to manage (e.g., split long strings).

cushon:
I understand that reformatting e.g. `//$NON-NLS-1$` to `// $NON-NLS-1$` (with a space after the `//`) is an issue for eclipse, and I think special-casing these comments similar to `//noinspection` is probably OK.

I'm not sure I understand the line wrapping issue, though: google-java-format doesn't currently reflow long string literals, or move line comments off very long lines.

e.g. this is not currently wrapped:

```java
class T {
  private String s =
      "long looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong string"; // $NON-NLS-1$
}

```

Do you have an example of a `//$NON-NLS-x$` comment where line wrapping is an issue?

cushon:
Thanks, is it an option to write that code like this instead?

```java
    } catch (CoreException ex) {
      logger.log(
          Level.WARNING,
          "Unable to obtain jst.web facet version from selected project", //$NON-NLS-1$
          ex);
    }
```

cushon:
[...]

PR #222, opened 2017-11-07
Update JavaCommentsHelper.java

romp13:
#221 this would solve the problem I believe


romp13:
I signed it!

cushon:
I asked a clarifying question in #221. I think I'd prefer special-casing `//$NON-NLS-x$`, the [style guide](https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace) doesn't make an exemption for line comments that don't contain whitespace.

PR #229, opened 2017-11-27
Moe sync

ronshapiro:


ronshapiro:
@cushon looks like we might have to bump the maven-javadoc-plugin version?

Issue #164, opened 2017-06-09
Incorrect formatting of then clause when there is a comment on the line with `if`


mernst:
Consider the following code, which has been formatted with google-java-format:

```java
public class CommentedTest {

  public boolean comparableOrSuperclassEitherWay(ProglangType other) {
    if (this == other) // ProglangType objects are interned
    return true;

    if (((this.base == BASE_OBJECT) && other.baseIsObject()) // interned strings
        || ((other.base == BASE_OBJECT) && baseIsObject())) // interned strings
    return true;

    return false;
  }
}
``` 

The two `return true;` lines ought to be indented, but they are not.

cushon:
Thanks for the report. There's a similar issue with:

```
return // foo
bar;
```

It's definitely a bug, just not one I've got around to fixing yet. (And the style guide discourages omitting the braces for if statements, so we don't have much code like your example.)

mernst:
I agree it's bad style.  If only such code didn't exist!  Since it does, thanks for having this on your to-do list.

PR #239, opened 2017-12-29
Fixed the formatting of statements with comments on the same line

kitonjicsm:
Fixed [#164](https://github.com/google/google-java-format/issues/164)

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/Doc.java
+++ b/core/src/main/java/com/google/googlejavaformat/Doc.java
@@ -603,6 +603,16 @@ public abstract class Doc {
     }
 
     /**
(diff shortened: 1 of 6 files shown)


cushon:
This change has some unexpected side-effects, for example:

Before:

```java
class T {
  @Deprecated // test
  public void f() {}

  int[] xs = {
    {1, 2, 3}, // test
    {4, 5, 6},
  };
}
```

After:

```java
class T {
  @Deprecated // test
  public
  void f() {}

  int[] xs = {
    {1, 2, 3}, // test
    {
      4, 5, 6
    },
  };
}
```

PR #249, opened 2018-01-24
Moe sync

ronshapiro:


idea_plugin/src/com/google/googlejavaformat/intellij/FormatterUtil.java
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      ImmutableMap.Builder<TextRange, String> replacements = ImmutableMap.builder();
+      formatter
+          .getFormatReplacements(text, toRanges(ranges))
+          .forEach(
+              replacement ->
+                  replacements.put(
+                      toTextRange(replacement.getReplaceRange()),
+                      replacement.getReplacementString()));
+      return replacements.build();
```
jbduncan: @plumpy I believe that the code in this try statement could be made more idiomatic with something like:

```java
return formatter
    .getFormatReplacements(text, toRanges(ranges))
    .stream()
    .collect(
        toImmutableMap(
            replacement -> toTextRange(replacement.getReplaceRange()),
            replacement -> replacement.getReplacementString()));
```

plumpy: This code is designed to run in IDEA 2017.1+, but 2017.1 has Guava 19, which does not have any Java 8 features like toImmutableMap.

plumpy: (and I'd happily bump to 2017.2 as a minimum version, but the current version of Android Studio is based on 2017.1, so I'd like to keep supporting that)

jbduncan: Ah I see, that's a bit of a shame. Thanks for the clarification. :)

PR #261, opened 2018-02-25
Add JDK 11 to Travis CI build matrix

sormuras:
Add `jdk-11-ea+[build]` to the Travis CI build matrix.

It fails to compile at the moment emitting the following exception:
```
[...]
[ERROR] /home/travis/build/google/google-java-format/core/src/main/java/com/google/googlejavaformat/java/DimensionHelpers.java:[18,17] cannot access java.util.ArrayDeque
  bad class file: /modules/java.base/java/util/ArrayDeque.class
    class file has wrong version 55.0, should be 53.0
[...]
```
I guess `errorprone` is not Java 11 compatible, yet.

JDK 8, 9, and 10 are "green". JDK 11 is allowed to be "red": https://travis-ci.org/google/google-java-format/builds/346019733


scripts/install-jdk.sh
```diff
@@ -0,0 +1,125 @@
...
+# it there.
+#
+# Example usage
+#
+#   install-jdk.sh                 | install most recent (early-access) JDK
+#   install-jdk.sh -W /usr/opt     | install most recent (early-access) JDK to /usr/opt
+#   install-jdk.sh -F 9            | install most recent OpenJDK 9
+#   install-jdk.sh -F 10           | install most recent OpenJDK 10
+#   install-jdk.sh -F 10 -L BCL    | install most recent OracleJDK 10
+#   install-jdk.sh -F 11           | install most recent OpenJDK 10
```
rtack: comment should be updated to OpenJDK 11 as well.

sormuras: Dooh. Done. Force-pushed changes.

sormuras:
With [JDK 10 General-Availability Release](http://jdk.java.net/10/) happened today, the current build configuration for Travis CI will fail. Merging this PR will prevent that failure. ;)

PR #273, opened 2018-04-12
Make google-java-format build with Bazel

cgrushko:
I don't expect this PR to be merged - it's just a PoC so we have something concrete to discuss.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ target/
 
 bin/
 out/
+
+bazel-*
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,3 @@
+load("//thirdparty:maven.bzl", "maven_dependencies")
+
+maven_dependencies()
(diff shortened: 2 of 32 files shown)


cgrushko:
@cushon

ronshapiro:
@cgrushko if we'd do this, I'd love to use github.com/google/bazel-common for all of the third-party deps. We're soon (cl/192520035 and cl/192645783) going to have maven pom.xml generation that takes advantage of those too, which I imagine we'd want here too.

cushon:
Thanks, this looks like a great start. As I mentioned elsewhere, some of the blockers for migrating off maven are:

* making our sync process handle `BUILD` files (and remap dependencies, etc.)
* figuring out how to handle releasing to maven

@ronshapiro it'd be great to reusing the solution you're using to handle maven releases for dagger.

jbduncan:
@cushon I'm not entirely sure why this PR was closed, so I wondered if you could explain it for me? :)

cushon:
It was a proof of concept for discussion, not something that was ready to merge.

jbduncan:
Fair enough, thanks for clarifying! :+1:

PR #286, opened 2018-05-31
Replace 'install-jdk.sh' by new jdk constants

sormuras:
Travis CI now uses 'install-jdk.sh' internally.

See https://github.com/travis-ci/travis-build/pull/1347


sormuras:
Adding `openjdk[-]`_xx_ to matrix is also possible. Even can replace the `oracle` ones.

Issue #266, opened 2018-03-16
shaded javac should set Automatic-Module-Name and shade service loader entries


sormuras:
When integrating `google-java-format` as a library (into an application that lives on the `module-path`) there are some issues. Especially with the 3rd-party dependencies `google-java-format` relies on.

One issue, namely the dependency to a certain `error-prone/javac` release, could be tackled by using `javac` provided by the current runtime system.

Copied from https://github.com/forax/pro/pull/34#issuecomment-373315521
_For the javac-9-dev-r3297-1-shaded.jar, it would be cool if google-java-format had an option to use javac through the service interface as we do in the compiler plugin instead of using its own version._

cushon:
Can you expand on what the issues are? The motivation for the shaded javac is that GJF uses internal javac APIs that change between releases. Supporting arbitrary javac versions doesn't seem that useful, because it will still only work with that latest version. What problems are you seeing with `-all-deps.jar` in that PR?

sormuras:
I'm just in the process of rebasing the PR to gjf 1.5 and current pro master. There are two ways to include gjf into an external application, like pro. a) Use the 3 single jars or b) use the `-all-deps.jar`.

a)
Quote: _I didn't managed to make modules via _module-fixer_ out of neither the "-all-deps" distro nor the 3 single jars:  `google-java-format-1.3.jar`, `guava-19.0.jar` and `javac-9-dev-r3297-1-shaded.jar`_

Not all problems I faced with the _none_ `-all-deps.jar` way come to my mind now.
Forax' pro tool runs entirely on the module-path. iirc, gjf poses no problem. Especially after providing an `Automatic-Module-Name`. Guava was not behaving well back then, and javac-... had some split-package issues.

b)
My solution then was to fallback to the command line and wrap/control the external process within a pro plugin. This worked then and does work now -- but it is not really ... integrated.

So, the idea is to implement the `formatter` plugin for pro using approach a). Pro always comes with a current/latest javac version bundled in its jlink-ed runtime image. It would be nice if gfj could use that javac via the tool interface first. Maybe gjf could introduce an option to use the default "javac" tool?

sormuras:
Did my explanation clarify the use case?

cushon:
Having an option to use other javacs is complicated by the way we're doing the shading. It's possible, but I'm not really understanding why it's necessary. It sounds like (a) works with `-all-deps` now?

sormuras:
[...]

PR #294, opened 2018-07-13
Moe Sync

ronshapiro:

Commits:
=====
<p> Use j.u.Optional in google-java-format

aa28143a8b515cf48bb2a1509f1c6022bd33dc87

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/Doc.java
+++ b/core/src/main/java/com/google/googlejavaformat/Doc.java
@@ -17,13 +17,13 @@ package com.google.googlejavaformat;
 import static com.google.common.collect.Iterables.getLast;
 
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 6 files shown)


sormuras:
...and next, replace guava utilities with java foundation helpers and uses `javac` via the compiler interface. :-)

cushon:
> replace guava utilities with java foundation helpers

Do you mean other JDK methods, or something else? I did `Optional` because the `Stream` methods return `j.u.Optional`s, and converting them is annoying. We aren't planning to stop using guava, but in places where the JDK now has equivalent helpers we'll probably slowly migrate.

> and uses javac via the compiler interface. :-)

Is this partly related to #266? The vendoring/shading approach is kind of unfortunate, but it seems less fragile than relying on internal APIs in `jdk.compiler`. What do you think?

Issue #288, opened 2018-06-05
Create official Docker image of CLI jar


matthewadams:
Please create an official Docker image so that the CLI can be used without having to download the CLI jar from the releases page (and, to a lesser extent, so that it can be used without having `java` installed).

aegixx:
Not official, but I pushed out a very basic one @ https://github.com/google/google-java-format/pull/303

aegixx:
@matthewadams As a not-that-simple one-liner too, I'm using this:
```
docker run -it --rm -v $(PWD):/src openjdk:alpine sh -c "wget https://github.com/google/google-java-format/releases/download/google-java-format-1.6/google-java-format-1.6-all-deps.jar && find /src -name *.java -exec java -jar google-java-format-1.6-all-deps.jar -i {} \;"
```

vandmo:
Feel free to copy stuff from https://github.com/vandmo/google-java-format
It has GitHub workflow for publishing to arm64 and amd64 which can easily be extended to more platforms.

PR #303, opened 2018-08-15
Very simple Dockerfile based on openjdk:alpine

aegixx:
Very basic, but easy implementation that could be pushed to a docker repo and utilized by others.


aegixx:
This potentially resolves https://github.com/google/google-java-format/issues/288

aegixx:
Any comments or resolution I can get on this?  @ronshapiro - Thank you!

Issue #306, opened 2018-09-05
Add support for IntelliJ 2018.3


tolbertam:
IntelliJ 2018.3 is currently in EAP, would be nice to update the plugin to support it.

tolbertam:
Haven't ran into any issues after ~6 hours of use so I think it should just work as is :+1:

vivekpatani:
How did you manage to install the plugin on IntelliJ 2018.3? The method mentioned on the README does not seem to the way to go it seems. Thanks.

plumpy:
@vivekpatani You can download the plugin, open the jar, edit plugin.xml and remove the "until-build" element. Or just clone the repository and modify idea_plugin/build.gradle like @tolbertam shows above.

I'll try to push a 2018.3 compatible version next week.

PR #309, opened 2018-09-29
Update the google-java-format plugin to support all Intellij 2018.x builds.

medb:
* update `org.jetbrains.intellij` Gradle plugin to latest version for compatibility with latest Gradle versions.
* mark Intellij plugin as compatible with all `2018.x` Intellij versions.

Fixes https://github.com/google/google-java-format/issues/306

### DIFF
--- a/idea_plugin/build.gradle
+++ b/idea_plugin/build.gradle
@@ -15,7 +15,7 @@
  */
 
 plugins {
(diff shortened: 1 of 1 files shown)


plumpy:
Thanks! I submitted a fix for this internally, which will make its way out to GitHub soon.

I also just uploaded a new version to the plugin repository.

PR #312, opened 2018-10-18
Fix broken JDK 10 Travis build

eaftan:


tbroyer:
The problem is the way Error Prone is added to the build.

Updating Error Prone to 2.3.2 (for JDK 10 and 11 support), and switching to `-Xplugin:ErrorProne` (see https://errorprone.info/docs/installation#maven), makes the project compile again.

But then it fails when generating javadocs due to `http://jsr-305.google.com/svn/trunk/javadoc/` no longer existing.

tbroyer:
See #313

cushon:
https://github.com/google/google-java-format/commit/92e0dfc00108b41b83d756849143ef0375d2142a

PR #315, opened 2018-10-25
Use replace-buffer-contents in google-java-format.el

sonologico:
This make it less jumpy, which makes it much nicer to use specially if you run it on the before save hook.


cushon:
I ran this by someone more familiar with emacs than I am, and they had this review feedback:

> you appear to be no longer trying to restore (point)
>
> It looks like this code is replacing the entire buffer, even though google-java-format was only asked to operate on a region.  Can you identify the region in the generated buffer containing the replacement text?  If so, it should be easy-ish to leave everything in the input buffer outside the selected region untouched (hint: save-restriction, narrow-to-region, replace-buffer-contents)

What do you think? If it's accurate that this is replacing the entire buffer, does the suggestion sound workable?

PR #313, opened 2018-10-22
Fix build with JDK 11

tbroyer:

PR #316, opened 2018-10-29
Moe Sync

### LINKED ISSUE #44: google-java-format-diff.py is unusable without rolling your own wrapper script
google-java-format-diff.py assumes there is an executable in $PATH named "google-java-format", which is not part of any published releases. The basic instructions say to run with "java -jar", so to make this work, someone has to write a wrapper script. I am capable of rolling my own but it really seems like this should be part of the distribution.

ronshapiro:

Commits:
=====
<p> Fix build with JDK 11

Fixes #313

e54321e0d59e33eedc2f0c497505364d85e50b29

-------

<p> Add flag to specify path to google-java-format binary in google-java-format-diff.py

Fixes #44

c5b565cce353020b400ec3b2dfdd48a200344d6a

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ jdk:
 
 matrix:
   allow_failures:
-    - jdk: openjdk11
     - jdk: openjdk-ea
 
 # see https://github.com/travis-ci/travis-ci/issues/8408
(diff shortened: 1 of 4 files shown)


tbroyer:
I confirm that I am ok with my commits being contributed to this project.

PR #321, opened 2018-11-14
Moe Sync

ronshapiro:

Commits:
=====
<p> Log the elapsed time to toolsearch

Also refactor the 'Main' class out of the library target, to avoid
introducing proto dependencies to users.

Roll forward of b383c5a14a7e594dfc7b45211eedc8ba50dd52c0

62b547ed916e0166b8cfabf14d2148bcb25dbd3c

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Toolsearch.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Toolsearch.java
@@ -1,0 +1,16 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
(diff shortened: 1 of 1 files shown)


jbduncan:
@cushon @ronshapiro I'm a bit confused by the commit introduced by this PR, as it introduces an empty class. Was this intended? 🤔

eaftan:
It wasn't intended.  It was supposed to be internal only, but Moe stripping apparently doesn't work the way we thought it does.  We'll fix in a followup.

PR #322, opened 2018-11-20
Moe Sync

ronshapiro:

Commits:
=====
<p> GoogleJavaFormat: Add --keep-going to java-format

Very useful when doing things like

google-java-format -i -k $(git diff --name-only)

And you have deleted files. Without the -k, it will fail completely.

a73c05ec0308a0b10b9aa21d8c22120ae82edfe1


cushon:
merged as 44308fe8d6364f4b31a4352782c4696dba138b2d

Issue #326, opened 2018-12-21
Upgrade Guava version to address vulnerability


xylo04:
The project currently depends on Guava 22.0, which is susceptible to CVE-2018-10237. The dependency should be upgraded to at least >24.1.

Context: I'm on the Apigee (now a Google) team and we're attempting to use google-java-format via Maven plugins. However, no plugin is going to be able to be imported into our artifact server while this vulnerability exists.

ronshapiro:
I'll send a change to update that. Though note that you can force maven to pick a different version by explicitly listing the version of Guava that you want.

xylo04:
Thanks for offering to make the change! I don't have any preference between my PR vs your change.

Re: forcing Maven, I can do that locally, but Apigee proxies all artifact fetches through Artifactory and has an import process with the aforementioned vulnerability scanner. I might be able to develop locally, but the CI server (Kokoro) wouldn't be able to access the plugin for verification.

xylo04:
@talios, your plugin is indeed the one we're looking at using. It's possible we could override the Guava version from within your plugin (I'm still learning about and getting comfortable with Maven's artifact version resolution). However, I think it's still worth fixing upstream (here) to hopefully benefit other integrators.

@ronshapiro, thanks for 5c07330! I assume it will take some manual work to create a new Maven release? No rush from my end.

ronshapiro:
@cushon is the right person to ask about releases, but yeah there is some button-pressing to get that to happen

talios:
@xylo04 I've released a new version of my Maven plugin (1.6.5) using Guava 27.0.1 and everything seems to run ok. But would also be good to get that merged upstream.

PR #327, opened 2018-12-21
Update dependency versions.

xylo04:
The only critical update here is the Guava version (issue #326). According to
Apigee's vulnerability tool, Guava 22 is vulnerable to CVE-2018-10237
"Mad Gadget" and must be upgraded beyond 24.1. Apigee won't be able to
use google-java-format via Maven until this is addressed.

The other version upgrades are nice-to-have.


xylo04:
Looks like the change is being made from inside Google; closing this PR in favor of that.

PR #337, opened 2019-01-27
Avoid potential NPE

Stephan202:
NB: This code is over a year old, so maybe the `null` branch is never taken?


cushon:
> maybe the `null` branch is never taken?

It seems that way, I think we may as well just remove the conditional instead. Thanks for pointing this out.

Stephan202:
The suggested alternative of simply dropping the `null` branch was merged in #367 (6930fca89f1d034a6bedfab1846489b61744ab32).

PR #347, opened 2019-03-08
Moe Sync

ronshapiro:

Commits:
=====
<p> Put empty class and member declarations on their own line

Per JLS 7.6 and 8.2 `;` is a valid empty top-level type or class member
declaration. This change puts it on its own line, instead of adding it
at the end of the preceding line.

Omitting the `;` is recommended: the goal of this change is just avoid a
partial formatting bug, and to call attention to the unnecessary element.

965411cdcb68e6b9abfcd3c023acbd19ce4190d8

-------

<p> Recognize atDebug methods in flogger heuristic

4cc87e48ef3fe13ccd8e9945db9ea2e38e996100

-------

<p> Add support for JetBrains 2019.1 IDEs.

225547d440f483f18bfd3e7dba8e3a675b4d8677

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -22,7 +22,7 @@
   <parent>
     <groupId>com.google.googlejavaformat</groupId>
     <artifactId>google-java-format-parent</artifactId>
-    <version>1.8-SNAPSHOT</version>
+    <version>1.7-SNAPSHOT</version>
   </parent>
 
   <artifactId>google-java-format</artifactId>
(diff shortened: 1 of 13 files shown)


core/pom.xml
```diff
@@ -22,7 +22,7 @@
   <parent>
     <groupId>com.google.googlejavaformat</groupId>
     <artifactId>google-java-format-parent</artifactId>
-    <version>1.8-SNAPSHOT</version>
+    <version>1.7-SNAPSHOT</version>
```
Stephan202: Here and below: the version downgrades seem accidental.

Issue #317, opened 2018-11-03
"Format code"  operation removes all existing breakpoints from formatted file in Eclipse IDE (2018-09)


eddi-weiss:
When I format whole java file, all existing breakpoints are removed from the file.

I am using eclipse IDE 2018-09 with google-java-format-eclipse-plugin_1.6.0.

jbduncan:
Hi @eddi-weiss, welcome to google-java-format!

I can't speak for the google-java-format team, but I'm struggling to understand what you mean by "breakpoints" in this context. Could you clarify for me? :thinking:

eddi-weiss:
Just create a debugger line breakpoint inside java editor. Then format the source code. 
-> The breakpont is removed. 

PS: see the movie 
[2018-11-04_00h44_15.zip](https://github.com/google/google-java-format/files/2545429/2018-11-04_00h44_15.zip)

jbduncan:
Ah, I see, thank you!

I wonder, does this problem reproduce if you format with Eclipse's built-in formatter and the google-java-format CLI tool?

I ask because if so, then it would suggest to me that it's an unavoidable part of how Eclipse works, but if not then it would suggest to me that it's a bug in the google-java-format Eclipse plugin.

tobous:
> I wonder, does this problem reproduce if you format with Eclipse's built-in formatter and the google-java-format CLI tool?
>
> I ask because if so, then it would suggest to me that it's an unavoidable part of how Eclipse works, but if not then it would suggest to me that it's a bug in the google-java-format Eclipse plugin.

This issue does not occur when using the built-in formatter (or at least I haven't encountered an edge-case where it does). It does, however, occur consistently when using the GJF plugin. This behavior seems to be caused by a combination of how Eclipse tracks breakpoints and how GJF formats code.

Eclipse breakpoints seem to be bound to the surrounding method as far as I can tell. As a result, the breakpoint is removed when the method is removed (even if just temporarily). This also applies when the method is just replaced.
This behavior can be observed by replacing editor content, either by using the replace functionality in Eclipse or by selecting and replacing text using copy&past. If only a part of the method is replaced (not including the method declaration), the breakpoint remains. If the whole method is replaced, the breakpoint is removed.

[...]

PR #371, opened 2019-05-08
fix breakpoint behavior

eddi-weiss:
Hi all,
It is not clear for me which part of code causes removing of breakpoints, but I think I fixed it.


eddi-weiss:
```I think this would regress support for Eclipse's F_INCLUDE_COMMENTS option.```
@cushon How can I test it? Can you provide a junit test?

cushon:
(https://github.com/google/google-java-format/issues/317)

PR #394, opened 2019-08-15
Moe Sync

netdpb:

Commits:
=====
<p> google-java-format: construct the Replacement range in the constructor, in order to guarantee it is closedOpen.

Remove the Replacement.create(Range,String) overload, since the bounds were not checked for correctness (open/closed, lower bound >= 0).

046b8a55f46362818105b7bf10332042a961a020

-------

<p> JavaInputAstVisitor: fix unchecked cast

(Warnings kept cropping up while compiling my StringWrapper changes...)

5501cfd2ea9cee8c8a77fb6c570ec20131d9fc10

-------

<p> Delete obsolete unused import logic

we no longer support removing imports that are used in javadoc.

a64fadcc7b27f1bb9fdadd796f554208ba6d5d2e

-------

<p> google-java-formatter: reformat portions of the text affected by reflowing

755d44167728317acf0272aba0c21c4041d5f6ea

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Replacement.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Replacement.java
@@ -14,6 +14,9 @@
 
 package com.google.googlejavaformat.java;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.collect.Range;
 import java.util.Objects;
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/google/googlejavaformat/java/Replacement.java
```diff
@@ -23,28 +26,20 @@
...
   public static Replacement create(int startPosition, int endPosition, String replaceWith) {
+    checkArgument(startPosition >= 0, "startPosition must be non-negative");
+    checkArgument(startPosition <= endPosition, "startPosition cannot be after endPosition");
     return new Replacement(Range.closedOpen(startPosition, endPosition), replaceWith);
   }
 
-  public static Replacement create(Range<Integer> range, String replaceWith) {
-    return new Replacement(range, replaceWith);
-  }
-
```
jbduncan: Hi @awturner and @netdpb, please correct me if I've missed something, but it looks to me that the removal of `Replacement.create(Range, String)` and the introduction of the `final` keyword is an API-backwards-incompatible change. Thus I'm wondering if either it should be reverted or if the next version of google-java-format should be "2.0" to indicate through SemVer rules that a breaking change has occurred.

What do you think? :)

PR #411, opened 2019-11-05
Make JavaOutput.flush() public

cgrushko:
so it can be accessed by other formatters.


cgrushko:
@cushon - what do you think?

PR #412, opened 2019-11-05
Pull JavaInput.{getkN,getToken} up to Input

cgrushko:
so `JavaOutput` can be reused in formatters of other languages.


cgrushko:
@cushon - what do you think?

PR #414, opened 2019-11-05
Add a note on executable google-java-format

suztomo:
Google-java-format-diff expects an executable `google-java-format` ([link](https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py#L101)), but this repository does not contain such file. This PR explains how to create such executable in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,14 @@ The formatter can act on whole files, on limited lines (`--lines`), on specific
 offsets (`--offset`), passing through to standard-out (default) or altered
 in-place (`--replace`).
 
+It is convenient to have a shell script named `google-java-format` with the
+java command:
+
+```
+#!/bin/sh
+java -jar /path/to/google-java-format-1.7-all-deps.jar "$@"
+```
+
 To reformat changed lines in a specific patch, use
 [`google-java-format-diff.py`](https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py).
 


suztomo:
@googlebot I fixed it.

cushon:
> Google-java-format-diff expects an executable google-java-format (link), but this repository does not contain such file

I think this was addressed in #44.

suztomo:
The issue 44 was addressed by adding flag to specify the location of the executable. Still this repository is missing how to create the executable.

cushon:
there's also https://github.com/google/google-java-format/commit/fa113c6df50fb97c5556c534eabe7b2745e7f993

suztomo:
That command line option lets users specify a custom JAR file, without using an executable. This does not describe how to create the executable.

cushon:
aren't the two approaches equivalent?

suztomo:
Yes the effect of two approaches (the custom JAR approach and the executable-in-PATH approach) is the same. This equivalence should not block enhancing the document for one approach. By the way, I like the executable-in-PATH approach because the command line parameter is shorter for google-java-format-diff.py.

suztomo:
@eaftan Thank you for checking. Would you merge this PR? I don’t have write access.

eaftan:
@suztomo I'm testing a new sync workflow.  By approving this PR, I had hoped it would be automatically pulled into an internal CL, which we could then approve and submit internally.  Then that change would be synced outward and this PR closed.  But approving this didn't create the internal CL.  Let me poke it and see if I can make it work.

Issue #243, opened 2018-01-09
Either require semicolons at the end of enum declarations, or don't


dborowitz:
Currently, gjf accepts either:

```java
enum Foo {
  BAR,
  BAZ;
}
```

or:

```java
enum Foo {
  BAR,
  BAZ
}
```

I had a code reviewer tell me to remove the semicolon. I really do not care one way or another, but I would rather not have to discuss it during code reviews. This is exactly what gjf is for, so it'd be nice to take a stand one way or the other.

JakeWharton:
don't forget
```java
enum Foo {
  BAR,
  BAZ,
  ;
}
```
which minimizes changed lines on add/remove.

cushon:
The formatter currently makes very few non-whitespace changes. (Fixing imports is the main example, it also reorders modifiers.) There are some technical reasons we've been slow to add more non-whitespaces fixes. Removing unnecessary semicolons is on the list of things we'd like to get to eventually.

dborowitz:
Point taken that this is different from a whitespace change.

That said, it this case is slightly different from purely unnecessary semicolons. Nobody is going to argue that putting ";;" at the end of every line is desirable, but people do, empirically, argue about this. Fixing ";;" would be more like correcting a typo, whereas this feels more like taking a stand on a should-be-trivial style decision.

PR #415, opened 2019-11-06
Support swallowing tokens

cgrushko:
For example, this allows eliminating superflous semicolons, e.g.,

```
class A{

};
```

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/OpsBuilder.java
+++ b/core/src/main/java/com/google/googlejavaformat/OpsBuilder.java
@@ -292,6 +292,17 @@ public final class OpsBuilder {
         /* breakAndIndentTrailingComment=  */ Optional.empty());
   }
 
+  /**
+   * If the next token matched the guessed token, consume it without expecting it to be in the
+   * output
+   */
+  public final void ignoreOptionalToken(String tokenToIgnore) {
+    Optional<String> nextToken = peekToken();
+    if (nextToken.isPresent() && nextToken.get().equals(tokenToIgnore)) {
+      tokenI++;
+    }
+  }
+
   public final void token(
       String token,
       Doc.Token.RealOrImaginary realOrImaginary,


cgrushko:
@cushon, what do you think? GJF doesn't remove these currently, but it does remove unused imports (I think?).

cushon:
This is something I'm interested in supporting (e.g. https://github.com/google/google-java-format/issues/243#issuecomment-356315049), but it's also appealing to me that the core formatting logic (unused import removal is a separate pass) only changes whitespace, since it makes it easier to verify that the formatter isn't ever making semantic changes.

Have you looked in to how this interacts with comment handling, and partial formatting? I think the logic there might assume the token stream doesn't change.

cgrushko:
Mmm... we don't have enough experience to say if it causes trouble. 

Re partial formatting - we don't support it. I always thought it's there to allow incremental migration to the formatter, which is a problem we hope to not have.

Re comment handling - what did you have in mind? 
Anyway, any kind of examples would be super appreciated :D 
Thanks!

cushon:
> Re partial formatting - we don't support it. I always thought it's there to allow incremental migration to the formatter, which is a problem we hope to not have.

That isn't fully how we view partial formatting--even if code is mass-reformatted initially, the implementation of the formatter will continue to change and improve, and reformatting everything every time the formatter changes creates churn with little benefit. Only reformatting changed lines avoids that churn. (We use the same approach with e.g. `clang-format`.)

> Re comment handling - what did you have in mind?
> Anyway, any kind of examples would be super appreciated :D

I haven't thought through this very carefully, but the core formatting logic just outputs (non-whitespace, non-comment) tokens, and then the implementation adds the comments back to the output. If a token is skipped, I'm not sure what would happen to the comments? I'd try examples with comments immediately before/after a skipped token and see what happens.

cgrushko:
Huh! you're totally right :) 

```
enum A {}; /* comment */ enum B(2)
```

The comment is dropped.

Thanks!

PR #423, opened 2019-12-12
fix trailing tabs at end of line comments

### LINKED ISSUE #422: Handling of trailing tabs in comments not idempotent
When formatting the following code

```
class Foo {
  void f() {
    int x = 0; // comment<TAB><TAB><TAB>
    return;
  }
}
```
where `<TAB>` represents ASCII 0x09 character the formatter removes the last tab character.
Thus when you apply the formatter again, it also removes the last tab.
Expected behavior would be either all trailing tabs removed or kept.

As Unicode reader discards last character if it is a whitespace preseving tabs would be difficult.

Eckaaaaaat:
Should fix #422

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/JavaOutput.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/JavaOutput.java
@@ -55,9 +55,9 @@ public final class JavaOutput extends Output {
   private final int kN; // The number of tokens or comments in the input, excluding the EOF.
   private int iLine = 0; // Closest corresponding line number on input.
   private int lastK = -1; // Last {@link Tok} index output.
-  private int spacesPending = 0;
   private int newlinesPending = 0;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/google/googlejavaformat/java/JavaOutput.java
```diff
@@ -387,7 +393,7 @@ public String toString() {
     return MoreObjects.toStringHelper(this)
         .add("iLine", iLine)
         .add("lastK", lastK)
-        .add("spacesPending", spacesPending)
+        .add("peningSpaces", spacesPending.toString().replace("\t", "\\t"))
```
eaftan: The first parameter should not have changed, right?  That is, it should be `"spacesPending"`.

Eckaaaaaat: Of course. Fixed that.

PR #432, opened 2020-01-07
MOE Sync 2020-01-07

### LINKED ISSUE #422: Handling of trailing tabs in comments not idempotent

netdpb:

Commits:
=====
<p> Remove trailing tabs from comments to make behavior idempotent.

Fixes #422, fixes #423

901608609c1584bf218315aabdeb078b7d22374f

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/JavaOutput.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/JavaOutput.java
@@ -18,6 +18,7 @@ import static java.util.Comparator.comparing;
 
 import com.google.common.base.CharMatcher;
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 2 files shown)


netdpb:
Adding CLA:yes because 9016086 comes from @Eckaaaaaat's PR #423.

Issue #436, opened 2020-01-26
Support Java > 11 language features


sormuras:
Afaict, GJF runs flawlessly on Java source code units up to [Java Language Specification](https://docs.oracle.com/javase/specs/) 11. That includes formatting `module-info.java` files.

- [x] _JDK 9 `module-info.java` handling_ **(Standard)** ➡️ #75 

With new official Java syntax being potentially introduced every six months, GJF fails on those files that already make use of new language features. Such as Switch Expressions, Text Blocks, or Records. Here's an excerpt of the [JDK 14 feature list](https://openjdk.java.net/projects/jdk/14/) which are relevant for GJF's goal "reformatting Java source code": 

- [ ] Pattern Matching for instanceof (Preview) [JEP 305](https://openjdk.java.net/jeps/305) ➡️ - 
- [ ] Records (Preview) [JEP 359](https://openjdk.java.net/jeps/359) ➡️ - 
- [ ] Switch Expressions **(Standard)** [JEP 361](https://openjdk.java.net/jeps/361) ➡️ #340
- [ ] Text Blocks (Second Preview, Standard with Java 15) [JEP 368](https://openjdk.java.net/jeps/368) ➡️ #382 

GJF should either support those features and format them or by-pass any "unknown blocks" automagically and format the rest of the source unit as good as possible.

Follow-up from: #387 _Support foundation `javac`_
Related thread on twitter: https://twitter.com/sormuras/status/1221515240854229000

cushon:
(Moving the discussion back here after some related comments in #443.)

We want to support new versions of the language, but (1) I think it's an open question what the best way to do that technically is, and (2) then it still a matter of prioritization and resources.

For (1),  I'm not seeing great alternatives to having one branch/release per JDK version, and e.g. the JDK 13 version of g-j-f would require having JDK >= 13. (We could conceivably use mrjars to package multiple of those releases together, at least for Java > 9, but I haven't decided yet if that solves more problems than it creates.)

Historically we've been using a shaded redistribution of javac so the formatter can depend on a consistent version of the internal APIs it uses, and making a best-effort attempt for that redistribution to be compatible with a wider range of JDKs than upstream javac is.

[...]

PR #443, opened 2020-03-16
MOE Sync 2020-03-16

cpovirk:

Commits:
=====
<p> Migrate from JSR-305 to the Checker Framework annotations

this is tangentially related to Java 11 preparedness.

dac42f8d35341ca952555aebedf349f62a29ba24

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -46,8 +46,8 @@
 
     <!-- Compile-time dependencies -->
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
+      <groupId>org.checkerframework</groupId>
+      <artifactId>checker-qual</artifactId>
       <optional>true</optional>
     </dependency>
     <dependency>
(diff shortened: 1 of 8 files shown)


sormuras:
> this is tangentially related to Java 11 preparedness.

Including supporting new syntax elements (Java language features) as collected in #436?

cpovirk:
Not sure what @cushon has in store, but CCing him in case he has anything to say.

sormuras:
Missed him at FOSDEM and the OCW in Brussels in early February.

cushon:
The language level changes in 11 (`var`, `private` interface methods) don't affect the parser/AST and should already be supported, and the features discussed in #436 were added or previewed after 11.

Supporting Java > 11 is still future work. In particular, there isn't an obvious way to do it without burning compatibility with Java <= 11 (which we aren't ready to do yet) or maintaining branches for each JDK release (which could happen but isn't free).

sormuras:
> [...] maintaining branches for each JDK release (which could happen but isn't free).

Copied from https://twitter.com/BrianGoetz/status/1221884070155735040

>I think you've still missed the point, which is: what you're wishing for, while reasonable, is far more costly than you're imagining.
> Absent a serious effort to find a sponsor willing to underwrite this expense (for many years), is there more here than "I want a pony"?

I still want that pony -- and am interested in helping out on that case.

PR #446, opened 2020-03-19
MOE Sync 2020-03-19

cpovirk:

Commits:
=====
<p> Use the built-in JDK 11 javac

instead of relying on the shaded/vendored version.

c769d701dacacd0359d70c2ced6331fab6b5b549


cpovirk:
I'm closing this PR, since I made some changes directly in the PR instead of making them internally first. Now that we know that those changes (which I got from @cushon) are enough to fix the failures with the initial PR, we'll make the fix internally, and I'll create and merge a new PR from it.

PR #447, opened 2020-03-19
MOE Sync 2020-03-19

cpovirk:

Commits:
=====
<p> Use the built-in JDK 11 javac

instead of relying on the shaded/vendored version.

5aeab725e057a5cdbdf451ec771b0a57e39d2ca7

-------

<p> Increase minimum required JDK version to 11

25a1a8011e492cc2d44d2ee859c253c97e8cd5fe


sormuras:
Yeah! ❤️

PR #454, opened 2020-03-31
Fix javadoc syntax

sormuras:
Fix javadoc syntax by wrap `<` into an inline code block.

```
[WARNING] ...\google-java-format\core\src\main\java\com\google\googlejavaformat\java\JavaInputAstVisitor.java:2577: warning - invalid usage of tag <
```


sormuras:
Something worked. 😉

eaftan:
I'm as surprised as you are. 😀

PR #465, opened 2020-04-29
Fix Eclipse plugin build

tobous:
Updates the Eclipse plugin build logic. The logic is now completely separate from the main google java format build. Instead of relying on the build artifacts of the core build, the needed google-java-format and guava build artifacts used for the build are now pulled from maven.

Updates google-java-format version used for the build to '1.9', the build JDK to 11, and removes old dependencies which are no longer used.

Updates tycho version to 1.7.0 in order to build with JDK > 8.

Updates the build guide in the README.

### DIFF
--- a/eclipse_plugin/pom.xml
+++ b/eclipse_plugin/pom.xml
@@ -14,27 +14,46 @@
   ~ limitations under the License.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <groupId>com.google.googlejavaformat</groupId>
-    <artifactId>google-java-format-parent</artifactId>
-    <version>1.6</version>
-  </parent>
 
+  <groupId>com.google.googlejavaformat</groupId>
   <artifactId>google-java-format-eclipse-plugin</artifactId>
-  <version>1.6.0</version>
   <packaging>eclipse-plugin</packaging>
-  <name>google-java-format Plugin for Eclipse 4.5+</name>
+  <version>1.8.0</version>
+
+  <name>Google Java Format Plugin for Eclipse 4.5+</name>
 
   <description>
-    A Java source code formatter that follows Google Java Style.
+    A Java source code formatter plugin for Eclipse that follows Google Java Style.
   </description>
 
   <properties>
-    <tycho-version>0.26.0</tycho-version>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+    <tycho-version>1.7.0</tycho-version>
+
+    <!-- specific library versions; must match declarations in build.properties and META-INF/MANIFEST.MF -->
+    <google-java-format.version>1.8</google-java-format.version>
+    <guava.version>28.1-jre</guava.version>
   </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.googlejavaformat</groupId>
+      <artifactId>google-java-format</artifactId>
+      <version>${google-java-format.version}</version>
+      <type>jar</type>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${guava.version}</version>
+      <type>jar</type>
+    </dependency>
+  </dependencies>
 
   <repositories>
     <repository>
@@ -44,16 +63,28 @@
     </repository>
   </repositories>
 
-  <dependencies>
-    <dependency>
-      <groupId>com.google.googlejavaformat</groupId>
-      <artifactId>google-java-format</artifactId>
-      <version>1.6</version>
-    </dependency>
-  </dependencies>
-
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>3.1.2</version>
+        <executions>
+          <execution>
+            <id>copy-dependencies</id>
+            <phase>initialize</phase>
+            <goals>
+              <goal>copy-dependencies</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <outputDirectory>lib</outputDirectory>
+          <includeScope>runtime</includeScope>
+          <excludeTransitive>true</excludeTransitive>
+          <overWriteSnapshots>true</overWriteSnapshots>
+        </configuration>
+      </plugin>
 
       <plugin>
         <groupId>org.eclipse.tycho</groupId>
@@ -61,7 +92,6 @@
         <version>${tycho-version}</version>
         <extensions>true</extensions>
       </plugin>
-
       <plugin>
         <groupId>org.eclipse.tycho</groupId>
         <artifactId>target-platform-configuration</artifactId>
@@ -97,7 +127,15 @@
         </configuration>
       </plugin>
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.7.0</version>
+        <configuration>
+          <source>11</source>
+          <target>11</target>
+        </configuration>
+      </plugin>
     </plugins>
   </build>
-
 </project>
--- a/eclipse_plugin/pom.xml
+++ b/eclipse_plugin/pom.xml
@@ -22,7 +22,7 @@
   <groupId>com.google.googlejavaformat</groupId>
   <artifactId>google-java-format-eclipse-plugin</artifactId>
   <packaging>eclipse-plugin</packaging>
-  <version>1.8.0</version>
+  <version>1.9.0</version>
 
   <name>Google Java Format Plugin for Eclipse 4.5+</name>
 
@@ -36,7 +36,7 @@
     <tycho-version>1.7.0</tycho-version>
 
     <!-- specific library versions; must match declarations in build.properties and META-INF/MANIFEST.MF -->
-    <google-java-format.version>1.8</google-java-format.version>
+    <google-java-format.version>1.9</google-java-format.version>
     <guava.version>28.1-jre</guava.version>
   </properties>
 
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -157,51 +157,6 @@
         </configuration>
       </plugin>
       <plugin>
-        <artifactId>maven-resources-plugin</artifactId>
-        <version>3.0.1</version>
-        <executions>
-          <execution>
-            <id>copy-resources</id>
-            <phase>package</phase>
-            <goals>
-              <goal>copy-resources</goal>
-            </goals>
-            <configuration>
-              <outputDirectory>../eclipse_plugin/lib</outputDirectory>
-              <resources>
-                <resource>
-                  <directory>target</directory>
-                  <include>${project.artifactId}-${project.version}.jar</include>
-                </resource>
-              </resources>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-dependency-plugin</artifactId>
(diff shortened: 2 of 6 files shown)


tobous:
This PR was initially created before the move to building with Java 11. After rebasing, the build fails due to the maven javadoc build with the following error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:jar (attach-javadocs) on project google-java-format-eclipse-plugin: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses packages in the unnamed module, but the packages defined in https://github.com/google/google-java-format/google-java-format/apidocs/ are in named modules.
[ERROR] 
[ERROR] Command line was: /lib/jvm/java-11-openjdk/bin/javadoc @options @packages
[ERROR] 
[ERROR] Refer to the generated Javadoc files in '/home/tobous/git/google-java-format/eclipse_plugin/target/apidocs' dir.
```

Even with the failure, the plugin jar is still build and seems to be completely functional. (And, if I remove the doc build, everything builds fine.)

I had a look at recent changes and found the adjustment to the core `pom.xml` made in https://github.com/google/google-java-format/commit/3c191c1326b77aabbf2a48c3edc19b628f4327d6#diff-357e4854869b2e21c38b1b437f11095a. I tried adding the same configuration to the `pom.xml` for the Eclipse plugin build without any success.

Any idea how to fix the javadoc build?

sormuras:
Move to JDK 15 - it contains a fix for https://bugs.openjdk.java.net/browse/JDK-8240169

tobous:
@googlebot I fixed it.

m273d15:
@googlebot I consent.

tobous:
> Move to JDK 15 - it contains a fix for https://bugs.openjdk.java.net/browse/JDK-8240169

@sormuras I am not really sure what you mean. As far as I can tell, the google java format core is also build using Java 11 and the javadoc build seems to work just fine. Only the javadoc build for the Eclipse plugin fails (i.e. if I disable the Eclipse build, the build passes).

Furthermore, wouldn't this set the minimal required Java version to 15 (which hasn't even been released yet)?

sormuras:
The underlying problem for why _"javadoc build for the Eclipse plugin fails"_ is due to the issue linked above. Upgrading to 15-ea could solve the problem, introduce other ones ... and is not an option today.

So, I guess, you have figure out the differences in how the core project and the Eclipse plugin configure their `javadoc` run.

tobous:
As far as I can tell, the javadoc run is configured in the `pom.xml` files. As mentioned, I tried copying the configuration section of the [core pom](https://github.com/google/google-java-format/blob/master/core/pom.xml#L80) to the pom of the Eclipse build but it did not fix the build. 
(Updating to the latest maven-javadoc-plugin release `3.2.0` did not help either.)

Would simply skipping the maven javadoc build for the eclipse plugin be an acceptable solution?

pom.xml
```diff
@@ -28,14 +28,14 @@
   <groupId>com.google.googlejavaformat</groupId>
   <artifactId>google-java-format-parent</artifactId>
   <packaging>pom</packaging>
-  <version>1.8-SNAPSHOT</version>
+  <version>${revision}</version>
```
tbroyer, line 31: This will break deployments to Maven repositories unless you throw one more plugin to it: https://maven.apache.org/maven-ci-friendly.html#install-deploy

eclipse_plugin/pom.xml
```diff
@@ -48,20 +49,28 @@
...
         <artifactId>tycho-maven-plugin</artifactId>
         <version>${tycho-version}</version>
         <extensions>true</extensions>
       </plugin>
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>11</source>
```
tbroyer, line 69: `<release>11</release>` would be better, particularly when building with more recent JDK versions as that guarantees that you won't also use JDK12+ APIs.

tobous: I copied this section from the current core [`pom.xml`](https://github.com/google/google-java-format/blob/master/core/pom.xml#L250).

README.md
```diff
@@ -65,6 +65,13 @@ The plugin adds a `google-java-format` formatter implementation that can be
 configured in `Window > Preferences > Java > Code Style > Formatter > Formatter
 Implementation`.
 
+#### Running with Java 11
+
+Starting with the release `1.8`, the `google-java-format` plugin requires Java 11 or later to run.
+
+As a result, you will have to run Eclipse with Java 11 or higher to use the plugin.
```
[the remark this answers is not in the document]
tobous: I only chose to explicitly address this in the REAMD as Eclipse does not report the plugin being incompatible with the local runtime environment to the user. Instead it just silently doesn't load the plugin.

Adding the same explanation for other IDE plugins seemed out of scope to me. But I can adjust the section to cover all IDEs or drop it entirely, whatever you prefer.

tobous:
Added logic to skip maven javadoc build for google java formatter Eclipse plugin.

m273d15:
> This PR addresses at least 4 separate things at once

True, I would:
* remove the whole "revision" handling (I just used it, because in my test setup I wanted to avoid inconsistent version numbers)
* reduce the PR to: "fixing the Eclipse plugin build (update Tycho version and configure m-compiler-p to target JDK 11)"
* adjust the maven set-up to build with the current 1.8 version (from maven central not the local build)
* remove the reference from the parent pom to eclipse (this would decouple the builds and you could move the eclipse build into a separate repo).

PhilippWendler:
Given that there recently was another release without an Eclipse plugin, it is really cool to see some improvements being worked on for the plugin build. As a user of google-java-format, I would like to say thanks!

Is there a chance that we get an official build of the plugin with google-java-format 1.8 after this gets merged?

tobous:
Reworked the PR updated the PR message to match the new content.
- reworked the build logic
- dropped the changes to the general version logic
- dropped the addition of the Java 11 README section
- dropped the change to always build the Eclipse plugin

I force-pushed the changes to clean up the commits of this PR. I hope this is not an issue.

The new build logic now works independently of the core build. The dependencies are pulled from maven instead of being provided by the core build. Local core snapshots can also be used for the build.

The change to always also build the Eclipse plugin as part of the core build has also been reverted as you did not seem to like it. This, however, also means that the new build logic is not covered by the CI.

tobous:
@sormuras @tbroyer Any feedback on the updated PR?

tobous:
@plumpy Are there issues with the PR? Is there something I can do to help it get reviewed? I would really appreciate any feedback.

jan-z:
The build works fine for me. Using the plugin on Eclipse 2020-03 without problems.

tobous:
It's been a while since anything happened on this PR. I still don't see any open issues with it but haven't gotten any further feedback. As far as I can tell, there also aren't any merge/compatibility issues caused by other changes that were merged in the meantime.

Is there a fundamental issue with the PR? The way I understand it from reading the issues regarding the Eclipse plugin, the plugin was not released for newer versions of the formatter as you did not have the time to update the build scripts. But, from the muted response/feedback I have gotten on this PR, it makes me wonder whether you have made the internal decision to drop the Eclipse plugin.

This would be sad to here as I was hoping to keep using the google java formatter for both IntelliJ IDEA and Eclipse, but it is still your decision to make. It would, however, be nice to get any response/an official statement on the Eclipse plugin situation to avoid people (pointlessly) investing time providing PRs or reporting bugs for it or asking questions about its status.

oliviercailloux:
Please anyone at Google (@tbroyer?), can you provide at least some statement about an intent to consider or not consider this PR? It would be good to know if this path has to be considered hopeless for us users of Eclipse, and time invested to search for another solution for formatting code systematically, or if we can keep some hope that this PR will be merged in the foreseeable future.

av1m:
@cushon I see you are the most active... You have no idea on the merge of this PR? Or could you just give us a feedback on the evolution of the eclipse plugin?

cristatus:
For those looking for a working plugin for the latest eclipse release, here is the one I built from @tobous's branch.

[google-java-format-eclipse-plugin-1.9.0.zip](https://github.com/google/google-java-format/files/5714790/google-java-format-eclipse-plugin-1.9.0.zip)

Extract the zip file and put the plugin jar inside the `dropins` folder.

gmasil:
I tried your fix and it works like a charm. Thank you very much!
Can be found [here](https://github.com/gmasil/google-java-format/tree/1.9c) with GitHub release, if someone is interested.

To the google guys: Please value his work with as little as clicking the merge button...

ParanoidUser:
> As far as I can tell, the javadoc run is configured in the `pom.xml` files. As mentioned, I tried copying the configuration section of the [core pom](https://github.com/google/google-java-format/blob/master/core/pom.xml#L80) to the pom of the Eclipse build but it did not fix the build.
> (Updating to the latest maven-javadoc-plugin release `3.2.0` did not help either.)
> 
> Would simply skipping the maven javadoc build for the eclipse plugin be an acceptable solution?
> 
> This can be easily done by adding `<maven.javadoc.skip>true</maven.javadoc.skip>` to the `properties` section of the Eclipse build pom.

Sorry for the late reply, I just joined. Looking at suggestions from @tobous and @sormuras - should we enforce a minimal JDK version that contains the Javadoc fix?

```
<plugin>
   <artifactId>maven-enforcer-plugin</artifactId>
   <executions>
      <execution>
         <id>enforce-versions</id>
         <goals>
            <goal>enforce</goal>
         </goals>
         <configuration>
            <rules>
               <requireJavaVersion>
                  <version>11.0.3</version> <!-- https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8212233 -->
               </requireJavaVersion>
            </rules>
         </configuration>
      </execution>
   </executions>
</plugin>
```

tobous:
> > As far as I can tell, the javadoc run is configured in the `pom.xml` files. As mentioned, I tried copying the configuration section of the [core pom](https://github.com/google/google-java-format/blob/master/core/pom.xml#L80) to the pom of the Eclipse build but it did not fix the build.
> > (Updating to the latest maven-javadoc-plugin release `3.2.0` did not help either.)
> > Would simply skipping the maven javadoc build for the eclipse plugin be an acceptable solution?
> > This can be easily done by adding `<maven.javadoc.skip>true</maven.javadoc.skip>` to the `properties` section of the Eclipse build pom.
> 
> Sorry for the late reply, I just joined. Looking at suggestions from @tobous and @sormuras - should we enforce a minimal JDK version that contains the Javadoc fix?
> 
> ```
> <plugin>
>    <artifactId>maven-enforcer-plugin</artifactId>
>    <executions>
>       <execution>
>          <id>enforce-versions</id>
>          <goals>
>             <goal>enforce</goal>
>          </goals>
>          <configuration>
>             <rules>
>                <requireJavaVersion>
>                   <version>11.0.3</version> 
>                </requireJavaVersion>
>             </rules>
>          </configuration>
>       </execution>
>    </executions>
> </plugin>
> ```

I separated the Eclipse plugin build from the main build a while ago. As a result, the javadoc job is no longer inherited from the main job, thereby avoiding the issue.

oliviercailloux:
For those interested in a solution to let Eclipse format code according to the Google style guide AND let Maven fail the build when the formatting is wrong, see [here](https://github.com/oliviercailloux/Relaxed-google-style) for a solution based on [Checkstyle](https://checkstyle.org/). Please report issues there if something goes wrong (I have only tested this on my install).

I hope it is not considered inappropriate to hijack this thread to suggest a workaround that has nothing to do with the Eclipse plugin to the Google formatting library. I hope that this comment may nonetheless be useful because I suspect that many readers of this page are, as I was, eager to solve the concrete code formatting problem, whether by this specific plugin or not.

cristatus:
I have just built the latest plugin for eclipse from @tobous's branch and rebased on official repo,

[google-java-format-eclipse-plugin-1.10.0.jar](https://github.com/cristatus/google-java-format/releases/download/1.10.0/google-java-format-eclipse-plugin-1.10.0.jar)

Put the plugin jar inside the `dropins` folder.

timgesekus:
Does anybody know, why this isn't merged?

cushon (approved):

cristatus:
Thanks for merging it. However, it required some updates for 1.10.0 release. I will make a pull request soon.

cristatus:
I just submitted a new pull request #635.

cushon:
Now that this and #635 is merged, I have uploaded a build of the Eclipse plugin to the releases page: https://github.com/google/google-java-format/releases/tag/v1.11.0

Sorry for the silence here. We're not using Eclipse as much these days, and I have limited expertise to support the plugin, but I will try to make sure that PRs to keep it working are merged in a more timely manner.

PR #485, opened 2020-05-19
Support quote in args-file(s)

### LINKED ISSUE #421: Fails to format files with a space in the path
I am running google-java-format on a Windows machine.  When I provide it a file (with the "@<file>" syntax) containing a list of files to format, it will fail to format files that contain a space in the path.  Example:

```
>type files.lst
C:\path\with\a space\Some.java

>java -jar google-java-format.jar @files.lst
Skipping non-Java file: C:\path\with\a
space\Some.java: could not read file: space\Some.java
```

Placing double-quotes around the path does not work, either.

It _does_, however, handle passing in the filenames in the command string, provided they are surrounded by double-quotes, e.g.:

```
>java -jar google-java-format.jar "C:\path\with\a space\Some.java"
```

aharonha:
Fixes #421 
To use this feature put this in your file:
```
@"C:\Users\User name\git\project\filename.jar"
```

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
@@ -14,9 +14,6 @@
 
 package com.google.googlejavaformat.java;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-
-import com.google.common.base.CharMatcher;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableRangeSet;
 import com.google.common.collect.Range;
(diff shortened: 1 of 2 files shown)


kfriedberger:
I just found this old PR #485 when creating a new PR for a similar topic: #931 .

Just a comment on this PR #485 here (I know it is quite old, but maybe someone reads it):
The change here solves only half of the issue, e.g., whitespaces in the parameter for the @-file itself are allowed.
However, this ignores any other parameter with whitespaces, like filenames for sourcefiles, and (from a user's perspective) it is unexpected that the quotes appear only after the @-symbol, like `@"some file"`, and not around the complete parameter, like `"@some file"` or `"some other param"`.

PR #489, opened 2020-05-27
Add support for yield statement

ntkoopman:
Implement visitYield. I copied the examples from [JEP 361](https://openjdk.java.net/jeps/361)


ntkoopman:
Sure, but note that `java14.input` also contains one example

PR #495, opened 2020-06-08
Support --skip-removing-unused-imports in google-java-format-diff.py

taesu82:


taesu82:
@googlebot I signed it!

Issue #538, opened 2020-10-24
Running on Java 16 fails


sormuras:
Running GJF `google-java-format-1.9-all-deps.jar` on Java 16 (`16-ea+21`) fails with the follow error
```
java.lang.NoSuchFieldError: reader
	at com.google.googlejavaformat.java.JavacTokens$CommentSavingTokenizer.processComment(JavacTokens.java:131)
	at jdk.compiler/com.sun.tools.javac.parser.JavaTokenizer.readToken(JavaTokenizer.java:949)
	at jdk.compiler/com.sun.tools.javac.parser.Scanner.nextToken(Scanner.java:115)
	at com.google.googlejavaformat.java.JavacTokens.getTokens(JavacTokens.java:88)
	at com.google.googlejavaformat.java.JavaInput.buildToks(JavaInput.java:364)
	at com.google.googlejavaformat.java.JavaInput.buildToks(JavaInput.java:334)
	at com.google.googlejavaformat.java.JavaInput.<init>(JavaInput.java:276)
	at com.google.googlejavaformat.java.Formatter.getFormatReplacements(Formatter.java:280)
	at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:267)
	at com.google.googlejavaformat.java.FormatFileCallable.call(FormatFileCallable.java:45)
	at com.google.googlejavaformat.java.FormatFileCallable.call(FormatFileCallable.java:26)
```

sormuras:
Due to

https://github.com/openjdk/jdk/commit/4f9a1ffcdd36f8c0cced9eeaa587fd439d4daf7b#diff-9dd71a3938fb726eb0eb9b7b30d8727167dad2402854b38464446c7a63ca1638L90-L94

the `JavaTokenizer` no longer has a `protected UnicodeReader reader` field but extends class `UnicodeReader` in the same package.

sormuras:
With #544 being merged, it'd be a good idea to add Java `ea` job to the test matrix (allowing it to fail) that test-drives the current early-access build of OpenJDK.

sormuras:
Preparing a PR here: https://github.com/sormuras/google-java-format/runs/1455621338 -- the `JDK Early-Access` job fails as expected with the error message shown in the issue description.

The overall workflow is marked as successful.
![image](https://user-images.githubusercontent.com/2319838/100278811-baf8c500-2f65-11eb-8604-ee4aefbad61e.png)

Although the `JDK Early-Access` job failed.
[...]

PR #548, opened 2020-11-25
Check build with JDK 16-ea

sormuras:
Addresses https://github.com/google/google-java-format/issues/538#issuecomment-733522202

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,31 @@ jobs:
         shell: bash
         run: mvn test -B
 
+  early_access:
+    name: 'JDK Early-Access'
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Check out repository'
+        uses: actions/checkout@v2
+      - name: 'Cache local Maven repository'
+        uses: actions/cache@v2
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -32,6 +32,31 @@ jobs:
         shell: bash
         run: mvn test -B
 
+  early_access:
+    name: 'JDK Early-Access'
+    continue-on-error: true
```
sormuras, line 37: This doesn't prevent the workflow from being marked as a failed. See https://github.com/actions/toolkit/issues/399 for details

sormuras: According to [continue-on-error](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)'s documentation, it should work... 🤔

cushon: Hmm. Is `continue-on-error` definitely supported for the entire `job`, or does it need to be on the affected `steps`? I saw a few examples of the latter looking at other github actions configs. Although, from https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/ it definitely sounds like it's supposed to work on the job.

sormuras: It did work as expected on the branch of my forked repository. See https://github.com/google/google-java-format/issues/538#issuecomment-733933661

Just this PR was marked as failed -- perhaps it would work after merging to master.

cushon: Ah, thanks for clarifying. I'll go ahead and merge this.

cushon (approved):

PR #557, opened 2021-01-04
Fixes before commit reformatting option for IntelliJ IDEA

### LINKED ISSUE #228: Before commit reformatting option in IntelliJ does invoke plugin
I have the IntelliJ plugin installed and reformatting works as expected. If I select 'Reformat code' option in commit window, then it will take my currently selected Code Style sheet and not use google-java-format.  It sets indentation to 4 spaces instead of 2 for example. Reformatting after committing then immediately gives local changes.

Is this a known issue? Will this be solved?

momenso:
Runs the plugin formatter instead of the default IDE Code Style formatting when "Reformat Code" is enabled in the Before Commit options.

Fixes #228.


plumpy (approved):
Thanks for figuring this out and sending the fix, I really appreciate it!

alexbde:
This is just great. In which release will it be rolled out?

Issue #540, opened 2020-10-31
Java 14 switch expressions with many values are badly formatted


abrudin:
When having many values covered by the same case, google-java-format does a terrible job formatting the code. Example:

```
public class GoogleJavaFormatFail {
  public String myMethod(final TestEnum testEnum) {
    return switch (testEnum) {
      case SOME_RATHER_LONG_NAME_1, SOME_RATHER_LONG_NAME_2, SOME_RATHER_LONG_NAME_3, SOME_RATHER_LONG_NAME_4, SOME_RATHER_LONG_NAME_5, SOME_RATHER_LONG_NAME_6, SOME_RATHER_LONG_NAME_7 -> "less than"
                                                                                                                                                                                                + " number"
                                                                                                                                                                                                + " 7";
      case SOME_RATHER_LONG_NAME_8 -> "number 8";
    };
  }

  public enum TestEnum {
    SOME_RATHER_LONG_NAME_1,
    SOME_RATHER_LONG_NAME_2,
    SOME_RATHER_LONG_NAME_3,
    SOME_RATHER_LONG_NAME_4,
    SOME_RATHER_LONG_NAME_5,
    SOME_RATHER_LONG_NAME_6,
    SOME_RATHER_LONG_NAME_7,
    SOME_RATHER_LONG_NAME_8
  }
}
```

This is a little sad, as the new switch statements are awesome for compile time type safety (make sure no case is missed).

PR #558, opened 2021-01-06
Wrap multiple values in switch expression case

fawind:
Closes https://github.com/google/google-java-format/issues/540

* Wrap multiple values covered in a switch expression case if they exceed the max line length.
  ```diff
  switch (e) {
  -  case SOME_RATHER_LONG_NAME_1, SOME_RATHER_LONG_NAME_2, SOME_RATHER_LONG_NAME_3 -> {}
  +  case SOME_RATHER_LONG_NAME_1,
  +    SOME_RATHER_LONG_NAME_2,
  +    SOME_RATHER_LONG_NAME_3 -> {}
  }
  ```
* Also collapse empty case blocks in switch expressions.
  ```diff
   switch (e) {
  -  case CASE_A -> {
  -  }
  +  case CASE_A -> {}
   }
  ```


abrudin:
Was it ever clear why this PR failed the Java 16 check? The logs have expired, is it possible to retrigger the checks for this PR?

fawind:
The failing JDK 16 and now 17 check seems to be unrelated to this change. I believe it's a general problem with the formatter as e.g. the same tests fail on master for the JDK 17 check ([ref](https://github.com/google/google-java-format/runs/2357076332)). I don't think it's in the scope of this PR to fix them.

@cushon, mind taking a look when you have a spare minute? Thanks!

cushon:
Thanks for the fix!

>  I don't think it's in the scope of this PR to fix them.

Agreed: the tests aren't passing on JDK 17 at head yet. That configuration is [marked as 'experimental'](https://github.com/google/google-java-format/blob/32b63aa2e4675cb03fb853ea84cbc37792e442a4/.github/workflows/ci.yml#L45) and won't block merging, but the UI makes it look like a failure.

Also I realized we're not breaking after the `->` in expression switches with expression bodies either, but that doesn't need to happen in this change:

```java
return switch (e) {
  case SOME_RATHER_LONG_NAME_1,
      SOME_RATHER_LONG_NAME_2,
      SOME_RATHER_LONG_NAME_3,
      SOME_RATHER_LONG_NAME_4,
      SOME_RATHER_LONG_NAME_5,
      SOME_RATHER_LONG_NAME_6,
      SOME_RATHER_LONG_NAME_7 ->
      fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff();
  case SOME_RATHER_LONG_NAME_8 ->
      fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff();
};
```

cushon (approved):

PR #627, opened 2021-07-19
Change deprecated Gradle command

mgenereu:
"compile" is now "compileOnly" in Gradle 7

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ configuration.
 
 ```groovy
 dependencies {
-  compile 'com.google.googlejavaformat:google-java-format:1.10.0'
+  compileOnly 'com.google.googlejavaformat:google-java-format:1.10.0'
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ configuration.
 
 ```groovy
 dependencies {
-  compileOnly 'com.google.googlejavaformat:google-java-format:1.10.0'
+  implementation 'com.google.googlejavaformat:google-java-format:1.10.0'
 }
 ```
 


mgenereu:
@googlebot I signed it!

README.md
```diff
@@ -121,7 +121,7 @@ configuration.
 
 ```groovy
 dependencies {
-  compile 'com.google.googlejavaformat:google-java-format:1.10.0'
+  compileOnly 'com.google.googlejavaformat:google-java-format:1.10.0'
```
jbduncan, line 124: I wonder if `implementation` is better than `compileOnly`. By my understanding, it ensures that google-java-format is present in the compiled JAR as well, which I imagine most people will expect. What do you think?

mgenereu: I was just doing a blind switch but I think `compileOnly` is right since it's not used at runtime by most implementations. By the time someone is using this at runtime, they should already know to tweak this to `implementation`.

jbduncan: I'm not sure I agree. The way I see it, this Gradle config and the Maven config above it should have the same behaviour to prevent confusion for people who know both tools.

But I'm happy to defer to you and the google-java-format team on this. :)

mgenereu: I didn't even look at the Maven entry and I don't know how to interpret that Maven entry.  I too defer to others on this.  Do I need to add a reviewer?

tbroyer: This is about using GJF as a library, so you'd want to use `implementation` (or `api` if the library exposes GFJ types).
If you're only using GJF as a tool to format a project's source code, then you wouldn't add it as a compile dependency; you'd either create a new Gradle configuration and use it as the classpath for a `JavaExec` task, or much more likely defer to the existing Gradle plugins that already support it (listed above).

tbroyer (changes requested):

plumpy (approved):

plumpy:
Oops, thanks to you too @jbduncan!

Issue #476, opened 2020-05-09
Need empty line before first member definition of a class


jjqq2013:
Thanks for your great work. I works well except some inconsistent output,

The output of following two java code are supposed to be same, but actually not, the both output the same as input.

```
public class Keys {
  private static class Identities0 {
    public static Identity<Long> IDENTITY_BILLING_OFFERINGS1 = Internal.create();
    public static Identity<Long> IDENTITY_BILLING_OFFERINGS2 = Internal.create();
  }
}
```
```
public class Keys {

  private static class Identities0 {

    public static Identity<Long> IDENTITY_BILLING_OFFERINGS1 = Internal.create();
    public static Identity<Long> IDENTITY_BILLING_OFFERINGS2 = Internal.create();
  }
}
```

I've tested this with version 1.8 and 1.7.

tbroyer:
The Google Java Style says:

> A blank line before the first member or initializer, or after the last member or initializer of the class, is neither encouraged nor discouraged.

(Source: https://google.github.io/styleguide/javaguide.html#s4.6.1-vertical-whitespace)

so it looks like google-java-format works as expected here, or at least it's not wrong.

kevinb9n:
We often still have the formatter standardize on a single choice anyway, but in this case we decided not to. There are big long hairy signatures for which it really helps to use a blank line after; whereas the second snippet above shows why a policy of always using the blank line is sad.

jjqq2013:
Thank you all for the comments. How do you consider the consistency? Essentially same java source files, generated difference output, though trivial?

kuchaguangjie:
It's ugly to have that 2 empty lines.

Issue #318, opened 2018-11-12
Advice requested: Respect existing line breaks


robfig:
I'm trying to get google-java-format to respect existing line breaks, as gofmt does. I want to make it  not insert or remove newlines in the middle of a statement, even if it adjusts indentation or other things about it.

I've spent some time going over the code, but I'm having trouble figuring out how to make such a change. Can you provide any pointers on how to accomplish this? 

Thank you!
-Rob

robfig:
Sorry, I didn't mean to propose to change this tool itself -- I wanted to make a fork of it that was more palatable to those used to gofmt. I believe without this change, I can't get the necessary buy-in to bring some relief to our Java developers that our Go developers have been enjoying. Gofmt has a similar approach and similar raves about it, so I don't think that my tweak is incompatible with what you pasted.

I had been hoping since it maintains interior blank lines within a function, that it might be feasible to accomplish if only I could be pointed in the right direction, but maybe that's not the case.

Thanks for responding!

kevinb9n:
Oh, I see. @cushon might have insights to help you, maybe. But it could also be that it's just too deeply baked in.

Issue #51, opened 2016-05-24
Should the formatter also add missing braces


aozarov:
To correct any violation of  https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

Currently it looks like a code that is formatted like
```java
if (condition)
  do_something
```

would be formatted (after running the tool) as:
```java
if (condition) do_someting
```

where as I would expect it to be formatted as:
```java
if (condition) {
  do_something
}
```

kevinb9n:
Yes, adding braces is a request we're considering (there are some concerns unique to making non-whitespace changes that we'd have to sort through).

In the meantime, the current behavior is acceptable, because (a) it's exactly what users following AOSP style want, and (b) for Google Style the code was already incorrect anyway, and maybe the odd change will prompt a human to make the right fix.

rjeberhard:
Any update on this?  I keep having to resolve flame wars between those who want what the Google style guide says vs. what the tools actually enforce.

Balaji94:
Will this be fixed anytime soon?

lazystone:
This feature would be really helpful. We have some legacy code where we have **a lot** of `if` conditions without braces. This makes it so error prone for a change. Just reformatting that code would improve readability and make it a bit friendly for changes.

kevinb9n:
Pasting comment from similar issue:

> So far, GJF has been focused on being a formatter, which is a somewhat narrower purpose than "style correcter". It does nothing but insert and remove whitespace characters, and sometimes reorder things. The more general tool would be useful, but I don't think we plan for this to become that. If cushon replies, take his word over mine. :-)

cushon:
+1 to what @kevinb9n said :)

FWIW we have an [Error Prone check for missing braces](http://errorprone.info/bugpattern/MissingBraces) that can be run as a refactoring, which we use internally to help enforce that style rule.

There are [a handful of other cleanups](https://github.com/google/google-java-format/labels/non-whitespace) like this that we may eventually be added to the formatter, but so far it hasn't been a priority, vs. focusing on pure whitespace reformatting changes that don't add or reorder tokens.

romani:
@cushon , 

https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

>Braces are used with if, else, for, do and while statements, even when the body is empty or contains only a single statement.
Other optional braces, such as those in a lambda expression, remain optional.

[...]

PR #629, opened 2021-07-23
Support sealed classes

### LINKED ISSUE #603: Sealed classes - error: expected token: 'sealed'; generated class instead
When trying to use GJF with JDK 16, I get the following error 

```
Caused by: com.google.googlejavaformat.java.FormatterException: 8:52: error: expected token: 'sealed'; generated class instead
        at com.google.googlejavaformat.java.Formatter.getFormatReplacements(Formatter.java:293)
        at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:267)
        at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:233)
        ... 135 more
```

With this source code

```java
package dev.zacsweers.moshix.sealed.sample.java;

import com.squareup.moshi.JsonClass;

import dev.zacsweers.moshix.sealed.annotations.TypeLabel;

public interface FailureTestCases {
  @JsonClass(generateAdapter = false, generator = "sealed:type")
  sealed class DuplicateLabels permits DuplicateLabels.TypeA, DuplicateLabels.TypeB {
    @TypeLabel(label = "a")
    static final class TypeA extends DuplicateLabels {}

    @TypeLabel(label = "a")
    static final class TypeB extends DuplicateLabels {}
  }

  @JsonClass(generateAdapter = false, generator = "sealed:type")
  sealed class DuplicateAlternateLabels permits DuplicateAlternateLabels.TypeA, DuplicateAlternateLabels.TypeB {
    @TypeLabel(label = "a", alternateLabels = {"aa"})
    static final class TypeA extends DuplicateAlternateLabels {}

    @TypeLabel(label = "b", alternateLabels = {"aa"})
    static final class TypeB extends DuplicateAlternateLabels {}
  }

  @JsonClass(generateAdapter = false, generator = "sealed:type")
  sealed class GenericSubtypes<T> permits GenericSubtypes.TypeA, GenericSubtypes.TypeB {
    // This form is ok
    @TypeLabel(label = "a", alternateLabels = {"aa"})
    static final class TypeA extends GenericSubtypes<String> {}

    // This form is not ok
    @TypeLabel(label = "b", alternateLabels = {"aa"})
    static final class TypeB<T> extends GenericSubtypes<T> {}
  }
}

```

I know sealed classes are still in preview though, so understand if this isn't expected to be supported yet. The error message is a little surprising though

now:
This implements support for JEP 409: Sealed Classes, see
https://openjdk.java.net/jeps/409.

The implementation handles the sealed and non-sealed keywords and sorts them
according to their JLS order.  This requires us to look at the actual text of
the Tok, as there’s no TokenKind for sealed and non-sealed.

The optional permits clause is handled in the same manner as the implements
clause. A new private method, classDeclarationTypeList, has been added to
facilitate this.

This fixes #603.

I implemented this in a manner that I deemed fit with the surrounding code, but if I missed something, don’t hesitate to get back to me and I’ll adjust this PR to suit.


now:
I just realized that I didn’t consider backwards compatibility at all.  I’ll add a Java15InputAstVisitor and handle it there.

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -154,19 +160,11 @@ public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOExcept
...
-                .asSubclass(JavaInputAstVisitor.class)
-                .getConstructor(OpsBuilder.class, int.class)
-                .newInstance(builder, options.indentationMultiplier());
-      } catch (ReflectiveOperationException e) {
-        throw new LinkageError(e.getMessage(), e);
-      }
-    } else {
-      visitor = new JavaInputAstVisitor(builder, options.indentationMultiplier());
-    }
+    if (getMajor() >= 15)
```
cushon, line 163: Our style guide requires optional braces (https://google.github.io/styleguide/javaguide.html#s4.1-braces), e.g.:

```diff
-if (getMajor() >= 15)
+if (getMajor() >= 15) {
```

Do you mind adding those, here and below? I can take care of it when merging the PR if that's easier.

core/src/main/java/com/google/googlejavaformat/java/java15/Java15ModifierOrderer.java
```diff
@@ -0,0 +1,31 @@
...
+import com.google.googlejavaformat.java.ModifierOrderer;
+import javax.lang.model.element.Modifier;
+
+public class Java15ModifierOrderer extends ModifierOrderer {
+
+  @Override
+  protected Modifier asModifier(Token token) {
+    Modifier m = super.asModifier(token);
+    if (m != null) return m;
+    else if (token.getTok().getText().equals("sealed")) return Modifier.valueOf("SEALED");
```
cushon, line 27: What do you think about putting this directly in `ModifierOrderer`? Checking the text of the modifier token and using `Modifier.valueOf` should still compile with earlier language levels. The main reason to have separate classes like `Java14InputAstVisitor` is to allow references APIs only available at the language level, like new AST nodes.

core/src/test/java/com/google/googlejavaformat/java/FormatterIntegrationTest.java
```diff
@@ -50,7 +50,7 @@
   private static final ImmutableSet<String> JAVA14_TESTS =
       ImmutableSet.of("I477", "Records", "RSLs", "Var", "ExpressionSwitch", "I574", "I594");
 
-  private static final ImmutableSet<String> JAVA16_TESTS = ImmutableSet.of("I588");
+  private static final ImmutableSet<String> JAVA16_TESTS = ImmutableSet.of("I588", "I603");
```
cushon, line 53: Can you add a `JAVA15_TESTS` entry for the new tests?

core/src/test/resources/com/google/googlejavaformat/java/testdata/I603.input
```diff
@@ -0,0 +1 @@
+sealed abstract class T permits D, E {}
```
cushon, line 1: Maybe add a couple more test cases, like one that has both a `permits` clause and `extends` or `implements`, and one where the signature doesn't all fit on one line?

cushon (changes requested):
Thanks for the contribution!

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
now: I can fix it, but I must first ask why google-java-format doesn’t format the source like this.  I ran all sources through google-java-format 1.10.0 before committing them.  Is it intended behavior to skip braces when there’s only one statement, or is that a bug in google-java-format?

(I see after running some tests of this that braces aren’t added, but will be retained if they’re there.  So this is intended behavior?  But I still wonder why the braces aren’t added if they’re missing.)

core/src/main/java/com/google/googlejavaformat/java/java15/Java15ModifierOrderer.java
now: Sorry, I should have mentioned my reasoning in my commit.  The thing that speaks against doing it that way is that `Modifier.valueOf` will throw `IllegalArgumentException` if given an unknown name, which will be the case for Java 14 and earlier.  I figured that this was a bit cleaner, as it avoids handling the potential `IllegalArgumentException` and also avoids handling (and dismissing) an exception for Java 15 and later, should there have been a typo in the source,  `"SELAED"`, for example.

core/src/test/java/com/google/googlejavaformat/java/FormatterIntegrationTest.java
now: That makes sense, I’ll fix that.

core/src/test/resources/com/google/googlejavaformat/java/testdata/I603.input
now: Yes, that makes sense.  It just struck me now that these tests will fail if not run on a JDK 15 or later.  Should I handle this in `core/pom.xml` and exclude them if that’s the case?

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
now: I just realized that I messed this up in another way.  I forgot to check for any exclusion rules in `core/pom.xml` and I see that `Java14InputAstVisitor.java` is being excluded there.  I’ll restore the code to load the visitor via reflection and make sure to set up the proper profile in `core/pom.xml` to handle `Java15InputAstVisitor.java` as well.

cushon: > I must first ask why google-java-format doesn’t format the source like this

See https://github.com/google/google-java-format/issues/51. The formatter's job is primarily to fix issues with whitespace, there are more rules in the style guide than just whitespace

core/src/test/resources/com/google/googlejavaformat/java/testdata/I603.input
cushon: The logic in `FormatterIntegrationTest.java` will ensure these tests are only run on JDK >= 15. They aren't compiled by maven it just sees them as resources. I don't think an exclusion in the pom is necessary.

core/src/main/java/com/google/googlejavaformat/java/java15/Java15ModifierOrderer.java
cushon: If there's an invalid modifier it should fail to parse and this code will never be executed, so I think it's fine to have it always try to handle `sealed` and `non-sealed`

core/src/test/resources/com/google/googlejavaformat/java/testdata/I603.input
now: Yes, that’s right, my mistake.

core/src/main/java/com/google/googlejavaformat/java/java15/Java15ModifierOrderer.java
now: Ah, of course.

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
now: I hadn’t realized that.  I’ll look into fixing #51 next ;-P.

now:
It’s unfortunate that `excludePackageNames` is a `String` and not a `String[]`.  The profile stuff gets a bit messy because of it.

I think I got everything set up correctly now, but if there’s still improvements to be made, I’ll gladly make them too :-).

core/src/main/java/com/google/googlejavaformat/java/java15/Java15InputAstVisitor.java
```diff
@@ -0,0 +1,34 @@
...
+public class Java15InputAstVisitor extends Java14InputAstVisitor {
+
+  public Java15InputAstVisitor(OpsBuilder builder, int indentMultiplier) {
+    super(builder, indentMultiplier);
+  }
+
+  @Override
+  protected List<? extends Tree> getPermitsClause(ClassTree node) {
+    return node.getPermitsClause();
+  }
```
cushon, line 33: One more thought here: we're going to need to add more version-specific visitors eventually, especially if there are new `visit*` methods for new AST nodes. However for cases where there's a new getter on an existing node, we're already using reflection in a couple of places, e.g. here: https://github.com/google/google-java-format/blob/bf5ef1c5de94fd623025e24b99ceadf18e83d8a5/core/src/main/java/com/google/googlejavaformat/java/java14/Java14InputAstVisitor.java#L60

I wonder if it'd be simpler to call `getPermitsClause` reflectively and defer adding a new visitor until there are new AST nodes, and to also clean up some of the reflection in `Java14InputAstVisitor` when that happens?

This is fairly subjective and I don't mind making that change when I import it, but let me know if you have any input on that.

core/src/main/java/com/google/googlejavaformat/java/ModifierOrderer.java
```diff
@@ -152,7 +152,15 @@ private static void addTrivia(StringBuilder replacement, ImmutableList<? extends
    * is not a modifier.
    */
   private static Modifier asModifier(Token token) {
-    return getModifier(((JavaInput.Tok) token.getTok()).kind());
```
cushon, line 155: WDYT about putting this in the `default:` in `getModifier`? This is currently the only call of `getModifier`, but putting the logic there would prevent a bug if another call of `getModifier` is added.

Also I'd consider using a switch instead of the if/else chain:

```java
switch (token.getTok().getText()) {
  case "sealed":
     return Modifier.valueOf("SEALED");
```

cushon (commented):
Thanks for the changes! This LGTM aside from two more nits

> If everything is A-OK, how would you like the commits? Should I squash everything and rewrite the commit message? Would you prefer that I squash it and break it up into smaller commits afterwards?

It'll get squashed when I merge it, the description in the PR message looks good as a commit message, I'll plan on using that as the message for the squashed commit

core/src/main/java/com/google/googlejavaformat/java/java15/Java15InputAstVisitor.java
now: I didn’t quite understand why these methods were being invoked reflectively in `Java14InputAsVisitor` to begin with.  The code is only compiled in a version that should have it.  Is it to handle the potential case of a preview feature getting changed/dropped in a future Java release?  Regarding `getPermitsClause`, that’s the final API in Java 17, as far as I can tell.  I don’t mind having it invoked reflectively in `Java15InputAstVisitor`, but I’d like to know the reason behind it (sorry, I should have asked earlier).

Or do you mean that we should perhaps drop `Java15InputAstVisitor` altogether and do the reflective call in `JavaInputAstVisitor` instead?  I’m fine with that too.  That’d simplify `pom.xml` and `Formatter`.  I implemented it this way to keep version-specific code out of `JavaInputAstVisitor` as much as possible, but perhaps this minor thing isn’t worth the effort.

core/src/main/java/com/google/googlejavaformat/java/ModifierOrderer.java
now: I didn’t do that, as `getModifier` takes a `TokenKind` (and returns early if that’s `null`), but I agree that it’s not optimal.  Would it be OK if I inline `getModifier` in `asModifier` and handle both cases using switches there?  (I agree that a switch is much better than an if/else chain.)

cushon: Inlining `getModifier` SGTM

core/src/main/java/com/google/googlejavaformat/java/java15/Java15InputAstVisitor.java
cushon: > Or do you mean that we should perhaps drop Java15InputAstVisitor altogether and do the reflective call in JavaInputAstVisitor instead?

Thanks, yes, I think I'd slightly prefer to drop `Java15InputAstVisitor` and use reflection in either `JavaInputAstVisitor` or `Java14InputAstVisitor`.

`getModule` is invoked reflectively because that method was added to `CompilationUnitTree` in JDK 17. In general the uses of reflection in `Java14InputAstVisitor` should all be for APIs that aren't the same for all versions >= 14 (because they were preview APIs that changed, or they were added in versions after 14).

My tentative plan for supporting future language versions is to introduce new `JavaNInputAstVisitor` classes for major versions that add new AST nodes and `visit*` methods, but to not necessarily add a new `Visitor` class for major versions if the only changes were in getters and it's not too unwieldy to use reflection.

now:
I think that that should do it.

One more question: It seems that there’s always an empty line after a class or interface declaration in this project, but I can’t find that in the style guide.  I don’t mind it being there, I’m only wondering what the reasoning for having it is.

cushon (approved):

cushon:
> One more question: It seems that there’s always an empty line after a class or interface declaration in this project, but I can’t find that in the style guide. I don’t mind it being there, I’m only wondering what the reasoning for having it is.

The style guide doesn't have a rule for that, there's some related discussion in https://github.com/google/google-java-format/issues/318#issuecomment-441764793 and https://github.com/google/google-java-format/issues/476#issuecomment-626220144

PR #637, opened 2021-07-31
Explicitly state GFJ version to use for Eclipse plugin build

tobous:
Re-adds the GJF version specification to the Eclipse plugin pom.

This decouples the GJF version from the Eclipse plugin version. As a
result, it is possible to increase the Eclipse plugin version without
changing the used GJF core version (e.g. in cases where you want to
increase the patch level of the plugin version when deploying fixes for
the Eclipse plugin).

Additionally, it allows for easier builds with local snapshot versions
of GJF.

Also updates the README to link to the current release. Removes explicit
mention of the Eclipse plugin release 1.6. Links to the release page
instead of a specific JAR file so that the link does not have to be
updated with every release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,9 @@ and import it into File→Settings→Editor→Code Style.
 
 ### Eclipse
 
-Version 1.6 of the
-[google-java-format Eclipse plugin](https://github.com/google/google-java-format/releases/download/google-java-format-1.6/google-java-format-eclipse-plugin_1.6.0.jar)
-can be downloaded from the releases page. Drop it into the Eclipse
+The current version of the google-java-format Eclipse plugin can be downloaded
+from the [releases page](https://github.com/google/google-java-format/releases)
+(`google-java-format-eclipse-plugin-X.jar`). Drop the jar into the Eclipse
 [drop-ins folder](http://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fmisc%2Fp2_dropins_format.html)
 to activate the plugin.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -70,9 +70,9 @@ and import it into File→Settings→Editor→Code Style.
 
 ### Eclipse
 
-Version 1.6 of the
-[google-java-format Eclipse plugin](https://github.com/google/google-java-format/releases/download/google-java-format-1.6/google-java-format-eclipse-plugin_1.6.0.jar)
-can be downloaded from the releases page. Drop it into the Eclipse
+The current version of the google-java-format Eclipse plugin can be downloaded
+from the [releases page](https://github.com/google/google-java-format/releases)
+(`google-java-format-eclipse-plugin-X.jar`). Drop the jar into the Eclipse
```
cushon, line 75: Thanks - these docs have already been updated in f2eabaa9a7a2fe1ccd17c30124d2041fe97a6313

cushon (changes requested):
> Additionally, it allows for easier builds with local snapshot versions of GJF.

Can you expand on the workflow you want to use with the `<version>` change? The checked-in versions of the poms are already using an unversions snapshot version: https://github.com/google/google-java-format/blob/00511536792cd6b72d3cfb83d426d25521874e80/pom.xml#L26, I think it should be possible to do snapshot builds of the plugin even without this change?

tobous:
> I think it should be possible to do snapshot builds of the plugin even without this change?

Yes, it is possible without the change. Maybe my commit message was misleading. I am not the biggest fan of using a single version argument defining both the version of the Eclipse plugin and the version of the used GFJ jar. This approach means that you always have to change the version of the Eclipse Plugin if you want to change the GJF version, e.g. for testing. But this is more a matter of taste.

The more important use case for separate version arguments is the possibility of independent releases for the Eclipse plugin (e.g. bugfix releases). The way I understand it, it is not possible to only increase the Eclipse plugin version with the current handling. If we were to set the plugin version to `1.11.1` to release an updated version of the plugin without switching GJF versions, the build would look for a GJF jar with the same version (which it would not be able to find).

PR #658, opened 2021-08-28
Since adopt doesn't exist anymore, update to zulu

geertjanw:


jbduncan:
`'temurin'` is the new name for `'adopt'` now. It's part of AdoptOpenJDK's renaming to [Adoptium](https://adoptium.net). So I'd argue that `'termurin'` is the right one to go for, but if it were me, I'd accept either that or `'zulu'`. :)

geertjanw:
@googlebot I signed it!

geertjanw:
> `'temurin'` is the new name for `'adopt'` now. It's part of AdoptOpenJDK's renaming to [Adoptium](https://adoptium.net). So I'd argue that `'termurin'` is the right one to go for, but if it were me, I'd accept either that or `'zulu'`. :)

Just be careful: it's not simply a change of name. Temurin is a great OpenJDK distro, it is part of the Eclipse Foundation (unlike Adopt) and TCK licensed, where Adopt was not (and probably could not really be called 'Java' as a result). However, Zulu has been consistent all along, anyway, full disclosure: I work for Azul, which produces Zulu, however, as someone who cares about Java, I'd be recommending Zulu anyway, for these reasons: https://www.azul.com/support-for-temurin/

cushon:
Thanks for raising this! I switched to zulu in 67f75cb5a5377346266a8ae93526c6bc5a7cb36c

(I was going to try `temerin`, but it doesn't seem to have a JDK 14 available yet, which we want to test against for now.)

PR #688, opened 2021-11-14
Update the IntelliJ plugin to use v1.12.0

ze:
Since the last plugin update was almost half a year ago, I thought it would be prudent to update it and resolve most of the issues regarding updating the plugin.

It took around half an hour once I managed to correctly set up the module due to the gradle plugin version 1.2.1 + intellij version 2020.3 requiring java 11 to develop.

All of the included annotation changes were requested by IntelliJ to add since the overrides added them as well.

The change of `<? extends TextRange>` to `<TextRange>` was to resolve the syntax errors introduced by a change removing the wildcard requirement.

NotificationGroup was deprecated and moved into the plugin xml file as per the SDK docs
The same was required for `StdFileTypes.JAVA` being deprecated with the fix being `JavaFileType.INSTANCE` + changing the depend module to add that fix.

I had to change the project SDK to JDK 11 on my end to have the gradle build work. 

I also tested the plugin with 1.12 and 1.10 on some test files such as from #653 and #654. Would be good to have some double checking though because I've had issues trying to get the test file reformatted in #558 working. Probably because it's locked behind an experimental flag?

### DIFF
--- a/idea_plugin/build.gradle
+++ b/idea_plugin/build.gradle
@@ -15,7 +15,7 @@
  */
 
 plugins {
-  id "org.jetbrains.intellij" version "1.3.0"
+  id "org.jetbrains.intellij" version "1.3.1"
 }
 
 repositories {
@@ -23,7 +23,7 @@ repositories {
 }
 
 ext {
-  googleJavaFormatVersion = '1.13.0'
+  googleJavaFormatVersion = "1.13.0"
 }
 
 apply plugin: "org.jetbrains.intellij"
@@ -35,7 +35,7 @@ targetCompatibility = JavaVersion.VERSION_11
 intellij {
   pluginName = "google-java-format"
   plugins = ["java"]
-  version = "2020.3"
+  version = "221.3427-EAP-CANDIDATE-SNAPSHOT"
 }
 
 patchPluginXml {
--- a/idea_plugin/build.gradle
+++ b/idea_plugin/build.gradle
@@ -26,20 +26,24 @@ ext {
   googleJavaFormatVersion = '1.13.0'
 }
 
-apply plugin: 'org.jetbrains.intellij'
-apply plugin: 'java'
+apply plugin: "org.jetbrains.intellij"
+apply plugin: "java"
+
+sourceCompatibility = JavaVersion.VERSION_11
+targetCompatibility = JavaVersion.VERSION_11
 
 intellij {
   pluginName = "google-java-format"
-  version = "IC-213.5744.18-EAP-SNAPSHOT"
+  plugins = ["java"]
+  version = "2020.3"
 }
 
 patchPluginXml {
   pluginDescription = "Formats source code using the google-java-format tool. This version of " +
                       "the plugin uses version ${googleJavaFormatVersion} of the tool."
-  version = "${googleJavaFormatVersion}.0"
-  sinceBuild = '201'
-  untilBuild = ''
+  version.set("${googleJavaFormatVersion}.0")
+  sinceBuild = "203"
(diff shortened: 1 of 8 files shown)


ze:
I see a new version released, I can update the PR to use 1.13 since it's really just a trivial version bump on Gradle.

idea_plugin/build.gradle
```diff
@@ -15,31 +15,35 @@
...
 }
 
 patchPluginXml {
   pluginDescription = "Formats source code using the google-java-format tool. This version of " +
                       "the plugin uses version ${googleJavaFormatVersion} of the tool."
-  version = "${googleJavaFormatVersion}.0"
-  sinceBuild = '201'
-  untilBuild = ''
+  version.set("${googleJavaFormatVersion}.0")
+  sinceBuild = "203"
```
plumpy, line 45: Curious why you updated this to 2020.3? AFAIK there's no reason it shouldn't continue to work with 2020.1. (I just tested the 1.13 plugin with 2020.1 and it still seems to work...)

idea_plugin/build.gradle
```diff
@@ -15,31 +15,35 @@
...
+apply plugin: "java"
+
+sourceCompatibility = JavaVersion.VERSION_11
+targetCompatibility = JavaVersion.VERSION_11
 
 intellij {
   pluginName = "google-java-format"
-  version = "211.6693.65-EAP-SNAPSHOT"
+  plugins = ["java"]
+  version = "2020.3"
```
plumpy, line 38: I like to build these against the latest version of IntelliJ possible. This causes a compile-time error if they added methods to `CodeStyleManager`, which they do from time to time (albeit not in a while now).

plumpy (commented):
Thanks, I just updated this internally to 1.13, so that will make it's way out to GitHub sometime next week (sorry for the delay).

But if you wouldn't mind rebasing onto that, I'd still love the rest of your changes! Thanks for taking the time to do this; I appreciate it!

plumpy:
...oh, actually, the 1.13 commit just landed, so you can rebase this now. Thanks again!

idea_plugin/build.gradle
ze: The reason I updated it to 2020.3 was for the newer features regarding notification groups
https://plugins.jetbrains.com/docs/intellij/notifications.html#top-level-notifications-balloons
\+ https://plugins.jetbrains.com/docs/marketplace/product-versions-in-use-statistics.html

I considered it acceptable to move to 203 from 201 to get the newer features to ease the pain in the future

ze:
Welp, more than a "few days or so" have passed, but I rebased. Perfectly willing to drop back down to `201` from `203`, but we will lose some of the newer gains which may become more of a problem further down the line

cushon:
Thanks for getting back to this!

No concerns from me, but I'll defer to @plumpy.

@ze did you see the comment about using the latest snapshot release here: https://github.com/google/google-java-format/pull/688#discussion_r755363817 ?

ze:
Yep! I updated it from `2020.3` to `221.3427-EAP-CANDIDATE-SNAPSHOT` which I think is the latest snapshot
source: https://www.jetbrains.com/intellij-repository/snapshots/

cushon (approved):

idea_plugin/build.gradle
plumpy: Oh okay, yeah that's totally reasonable. I thought it was just an arbitrary change, but if it's enabling that notification configuration change, that defintiely makes sense.

plumpy:
Thanks again @ze, I appreciate your help here!

PR #728, opened 2022-01-24
GraalVM native-image Maven build support

### LINKED ISSUE #358: Add native image builds
It would be awesome to produce a native binary build using graalvm's native-image as part of the release.

native-image seems to work well w/ the current all-deps jar.
```
native-image -jar google-java-format-1.7-all-deps.jar
```

perezd:
Closes #358.

This appears to work as expected. I have a [weird case](https://github.com/oracle/graal/discussions/4254) I'm not really sure how to address (the need to specify -Djava.home during execution), but otherwise, this produces a fully static binary that is very fast and works.

To execute the build, you run the following command:

```
mvn -Pnative -DskipTests package
```

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -280,9 +280,9 @@
                 <param>${project.build.directory}/${project.artifactId}-${project.version}-all-deps.jar</param>
               </classpath>
               <buildArgs>
-                <buildArg>-H:IncludeResourceBundles=com.sun.tools.javac.resources.javac</buildArg>
-                <buildArg>--verbose</buildArg>
+                <buildArg>-H:IncludeResourceBundles=com.sun.tools.javac.resources.compiler</buildArg>
                 <buildArg>--no-fallback</buildArg>
+                <buildArg>--static</buildArg>
               </buildArgs>
             </configuration>
           </plugin>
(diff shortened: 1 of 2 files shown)


cushon (approved):

nhoughto:
The simplest (only?) workaround for the `java.home` thing would be to add a new wrapping native-only class exposing a main(String[] args) entry method, that just does a dumb System.setProperty(..) and then calls the normal main entry method. Minimal change and would then work straight out of the box, it would be nice if you could specify some runtime args at native image compile time but you can't it seems. For example even heap arguments (-Xmx and friends) can't be defaulted which can be a pain.

cushon:
For `java.home` the code doesn't actually rely on the value at runtime, so we can also use `--initialize-at-build-time=` as a workaround to prevent the crash: https://github.com/oracle/graal/discussions/4254#discussioncomment-2070904

> For example even heap arguments (-Xmx and friends) can't be defaulted which can be a pain.

I just recently learned that as of 20.0 there's are flags like `-R:MaxHeapSize` that allow sizing the heap at image build time: https://www.graalvm.org/22.0/reference-manual/native-image/MemoryManagement/#java-heap-size

liuchong:
Great job! 👍

PR #732, opened 2022-01-28
Automate releases w/ GitHub Actions

perezd:
Attempting to fully automate the release process here, this first attempt is focused on the initial build infrastructure.
Additional work I wanna do:

- Templatize README.md so we can just find/replace the version numbers easy.
- Actually push the signed jars to sonatype
- Begin developing the graalvm/native image aspects.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,10 +8,10 @@ on:
         required: true
 
 jobs:
-  define-release:
+  build-maven-jars:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: read
     steps:
       - name: Setup Signing Key
         run: |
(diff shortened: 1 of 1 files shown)


perezd:
Additional work I wanna do:
- Templatize README.md so we can just find/replace the version numbers easy.
- Actually push the signed jars to sonatype
- Begin developing the graalvm/native image aspects.

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+        env:
+          NEW_VERSION: ${{ github.events.input.version }}
+        run: |
+          mvn versions:set versions:commit -DnewVersion="${NEW_VERSION}"
+          git ls-files | grep 'pom.xml$' | xargs git add
+          git commit -m "Release google-java-format ${NEW_VERSION}"
+          git push
+
+      - name: Build Jars
+        run: mvn clean verify gpg:sign -DskipTests=true -Dgpg.passphrase="${{ secrets.GPG_PASSPHRASE }}"
```
cushon, line 41: I think once we're doing the actual `mvn deploy` it will produce the same local jars to we probably don't need tow separate checks for this. Also we should run the tests on at least one of the invocations (by omitting the `-DskipTests=true`), but this will be faster for manually testing the workflow

cushon (approved):

.github/workflows/release.yml
perezd: SGTM

PR #758, opened 2022-03-07
Use burningwave lib to export javac packages from module

klaraward:
Just adding this as a draft PR to start a discussion.


klaraward:
Replaced by #759 
(Which doesn't help for things uses this lib as an API, for example fmt-maven-plugin, trying the same approach there - https://github.com/spotify/fmt-maven-plugin/pull/139)

Issue #5, opened 2015-05-16
Handle type annotations


cushon:
We always want to format type annotations in-line:

```
@Nullable Object foo() {}
```

This is tricky in contexts where type annotations could be mixed together with declaration annotations, such as field, method, and variable declarations. There's no syntactic information to we can use to distinguish between `@NullableType Object foo() {}` and `@Deprecated Object foo() {}`, but we want to output the first annotation in-line, and the second vertically.

Currently, we output ambiguous annotations as declaration annotations. This is fine for Java 7, and has the advantage of never incorrectly formatting a declaration annotation as a type annotation.

Here are some possible heuristics we could use to support this case for Java 8:

1) If the annotation occurs before a regular modifier, it's probably a declaration annotation:

```
@Deprecated
static Object foo() {}
```

2) If the annotation occurs after regular modifiers, and it isn't on a void-returning method, it's probably a type annotation:

```
static @Nullable Object foo() {}
```

3) If the declaration isn't a void-returning method, and it has no regular modifiers, consider the existing formatting: type annotations are probably on the same line as the type, declaration annotations are probably vertical. (The risk here is that we'd preserve things like `@Override Object foo() {}`.)

4) If the declaration is named `@Override`, it's probably a declaration annotation. This would offset the disadvantage of (3), but it could be confusing if `@Override` behaves differently than other annotations.

5) An annotation with parameters is unlikely to be a type annotation, and should probably be formatted vertically.

cpovirk:
>
> 5) An annotation with parameters is unlikely to be a type annotation, and
> should probably be formatted vertically.
>
Will DI binding annotations like @Named("foo") ever be considered to be
type annotations? (I assume that they're not today, since Guice+Dagger
don't require Java 8).

kevinb9n:
imho, they *could* be, but it seems awfully dodgy.

cgdecker:
If nothing else, changing binding annotations to type annotations seems unlikely/impossible since it would be an incompatible change to the JSR-330 spec. That said, I don't see any particular reason to think that type annotations would be especially likely to not have parameters.

mernst:
The proposed heuristics are generally reasonable.  However, I am concerned about this one:

> 5) An annotation with parameters is unlikely to be a type annotation

[...]

PR #802, opened 2022-07-15
Expand hard-coded list of type qualifiers

mernst:
This change is necessary even if a `-Dtypeannotationconfig=_filename_` flag is implemented as mentioned in https://github.com/google/google-java-format/issues/5.  (That filename would be for recently-created annotations, or ones that are private to a project or a company.)


mernst:
The CI failure is the same as is currently occurring on the master branch.

cushon:
> This change is necessary even if a -Dtypeannotationconfig=_filename_ flag is implemented as mentioned in https://github.com/google/google-java-format/issues/5

I think I see what you mean, but that isn't the way I was thinking about it. With `-Dtypeannotationconfig=` people could create their own lists, and we wouldn't have to try to maintain an exhaustive list of type use annotations in public libraries.

mernst:
> > This change is necessary even if a -Dtypeannotationconfig=_filename_ flag is implemented as mentioned in #5
> 
> I think I see what you mean, but that isn't the way I was thinking about it. With `-Dtypeannotationconfig=` people could create their own lists, and we wouldn't have to try to maintain an exhaustive list of type use annotations in public libraries.

Thanks for clarifying how you were thinking about it; that was not clear to me before.  My view is (as stated in the PR description) that the extra filename would be rarely used:  for recently-created annotations or for ones that are private to a project or a company.

If we apply your view consistently, then GJF should not support `org.jspecify.nullness.Nullable`.  Every user who wishes to use that annotation should create a side file containing its name and configure GJF to use that file.  I think this is unreasonable; do you agree?

There is no need to maintain an *exhaustive* list -- no one (other than you?) has suggested that.  But maintaining a *useful* list would be helpful to users.  It would significantly reduce user effort, and it would reduce inadvertent formatting differences among users and among projects.

Another argument in favor of this pull request is that it is consistent with how the current GJF implementation works.

cushon:
This is not the approach I want to take at this time. I would take a PR for the approach described in https://github.com/google/google-java-format/issues/5#issuecomment-1183638329

mernst:
I would be curious to understand the reasoning.  I noted some tradeoffs that favor this pull request, but you didn't have any criticism of it, nor mention anything that recommends the approach of every user having to create a side file and customize their buildfile.

Issue #94, opened 2016-11-17
IntelliJ plugin should sort imports in group alphabetically. But the plugin do not sort.


diguage:
Google Java Style should sort imports in group alphabetically. But the plugin do not sort.

It is as follows in my IDEA:

```java
import javax.annotation.Resource;
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletResponse;
import java.text.DateFormat;
import java.text.ParseException;
import java.text.SimpleDateFormat;
import java.util.Calendar;
import java.util.Date;
import java.util.List;
```

It should be as follows:

```java
import java.text.DateFormat;
import java.text.ParseException;
import java.text.SimpleDateFormat;
import java.util.Calendar;
import java.util.Date;
import java.util.List;
import javax.annotation.Resource;
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletResponse;
```

sormuras:
I think, the IDEA plugin uses the `Formatter.format()` entry point, which does _not_ optimize/fix (remove unused and sort) import statements. See #92 for further details.

thibaut-sticky:
Any update on this topic? I cannot find any call to Formatter.format() in the plugin package.

dwb5013:
Hi @plumpy any update on this?

reda-alaoui:
I'd like to create a pull request for this.

Parsing the plugin source code lead to general questions about its architecture. The plugin was "revived" in 2016. Before 2016, the plugin was already extending `com.intellij.psi.codeStyle.CodeStyleManager`. IntelliJ default implementation of `CodeStyleManager` abstract class delegates to interfaces `FormattingService` and `PostFormatProcessor`. 

According to the javadoc, those interfaces are more suited as extension points for formatting plugins. For example, when a plugin wants to provide an import formatting feature, it is expected to implement `PostFormatProcessor`.

I suppose `FormattingService` and `PostFormatProcessor` didn't exist at the time of plugin inception. Unless there is something I missed, wouldn't it make sense to start by restructuring the plugin to implement `FormattingService` instead of extending `com.intellij.psi.codeStyle.CodeStyleManager` ? Otherwise, could someone tell me why overriding `CodeStyleManager` is the best way to do it?

reda-alaoui:
#820 replaces `CodeStyleManager` "hack" with `FormattingService` implementation.

plumpy:
Version 1.16.0.0 includes an import optimizer to fix this issue.

PR #820, opened 2022-08-20
Stop relying on IntelliJ internals

reda-alaoui:
As of today, it looks like there is no more need to "hack" `CodeStyleManager`.

With this change, the plugin does not override `CodeStyleManager` anymore neither relies on `ComponentManagerImpl` IntelliJ internal class. Instead, the plugin extends the `FormatterService` interface which is the official Jetbrains way of providing formatting features.

`FormatterService`  opens up the ability to provide an `ImportOptimizer`. If this change gets merged, I will open another change to add import reordering ( #94 ).


reda-alaoui:
Let alone, `CodeStyleManagerImpl` executes `removeEndingWhiteSpaceFromEachRange` before calling `FormatterService`. I guess that's why we replace `CodeStyleManager` instead of extending of downstream extension points.
Closing this.

reda-alaoui:
I changed my mind (again :) ) => #821

PR #847, opened 2022-10-15
Fix 846 Formatter leaks threads and memory

stiemannkj1:
I've signed the CLA.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Main.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Main.java
@@ -18,6 +18,7 @@ import static java.lang.Math.min;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.common.io.ByteStreams;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.googlejavaformat.FormatterDiagnostic;
 import com.google.googlejavaformat.java.JavaFormatterOptions.Style;
 import java.io.IOError;
@@ -35,6 +36,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 /** The main class for the Java formatter CLI. */
 public final class Main {
@@ -187,6 +189,19 @@ public final class Main {
         outWriter.write(formatted);
       }
     }
+
+    // #846 Clean up any resources and threads created by the executorService.
+    final boolean completedShutdown =
+        MoreExecutors.shutdownAndAwaitTermination(executorService, 100, TimeUnit.MILLISECONDS);
+
+    if (!completedShutdown) {
+      // We wait for all formatting tasks to complete when looping through the formatting tasks
+      // before shutdown, so no tasks should be left over by the time we shut down the executor
+      // service.
+      throw new IllegalStateException(
+          "Failed to complete all formatting tasks and shut down the formatter.");
+    }
+
     return allOk ? 0 : 1;
   }
 


core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -186,6 +186,17 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
         outWriter.write(formatted);
       }
     }
+
+    // #846 Clean up any resources and threads created by the executorService.
+    int incompleteFormattingTasks = executorService.shutdownNow().size();
```
cushon, line 191: Let's use [`MoreExecutors.shutdownAndAwaitTermination`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/MoreExecutors.html#shutdownAndAwaitTermination(java.util.concurrent.ExecutorService,java.time.Duration)) here

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
```diff
@@ -613,4 +614,49 @@ public void noFormatJavadoc() throws Exception {
...
+              testB.toString());
+      assertWithMessage("Error Code").that(errorCode).isEqualTo(0);
+    }
+
+    // Verify that the number of threads is nearly the same before and after. NB: we allow for a few
+    // additional threads since we don't completely control the number of running threads in this
+    // test. In the failing case there are ~100 extra threads, so this should be sufficient to catch
+    // bugs.
+    assertThat(Thread.currentThread().getThreadGroup().activeCount())
+        .isLessThan(numberOfThreadsBeforeFormatting + 5);
```
cushon, line 660: Thanks for taking the time to write a test here. This seems hard to test in a unit test that's quick to run and non-flaky, though--I'm thinking of the comment about 10k iterations, and checking for the number of threads `+-5`. I think I'd be OK with just skipping a test for this.

cushon (changes requested):
Thanks for the fix!

core/src/test/java/com/google/googlejavaformat/java/MainTest.java
stiemannkj1: Removed.

core/src/main/java/com/google/googlejavaformat/java/Main.java
stiemannkj1: Done although this is a weird use-case for that method since it always requires a timeout. I choose 100ms because the method doesn't allow a timeout of 0. But at this point in the code we have already waited for each task to complete individually in the loop above.

cushon (approved):

stiemannkj1:
@cushon do I need to do anything to get the `import/copybara` workflow to complete?

cushon:
> @cushon do I need to do anything to get the `import/copybara` workflow to complete?

Nope, thanks! I can take it from here

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -187,6 +189,19 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
         outWriter.write(formatted);
       }
     }
+
+    // #846 Clean up any resources and threads created by the executorService.
+    final boolean completedShutdown =
+        MoreExecutors.shutdownAndAwaitTermination(executorService, 100, TimeUnit.MILLISECONDS);
```
cushon, line 195: I'm thinking to use a slightly longer deadline here, maybe 5 seconds.

This is just an FYI--I can make the change, no need to upload another snapshot.

core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -187,6 +189,19 @@ private int formatFiles(CommandLineOptions parameters, JavaFormatterOptions opti
...
+
+    if (!completedShutdown) {
+      // We wait for all formatting tasks to complete when looping through the formatting tasks
+      // before shutdown, so no tasks should be left over by the time we shut down the executor
+      // service.
+      throw new IllegalStateException(
+          "Failed to complete all formatting tasks and shut down the formatter.");
```
cushon, line 202: I think this should print the message to stderr and return with a non-zero exit code, similar to the error handling on other code paths here.

PR #848, opened 2022-10-19
Return a non-zero exit code upon error

abstractlyZach:
This brings us in line with the Unix standard of returning non-zero for a failed command:
https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html

Also, this will allow any callers (like google-java-format-diff.py) to respond appropriately. 
[A change](http://r.android.com/2257232) needed to be reverted because `google-java-format-diff.py` 
was failing silently, so the author wasn't aware that their change caused a pre-upload hook to stop working.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Main.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Main.java
@@ -79,7 +79,8 @@ public final class Main {
       return formatter.format(args);
     } catch (UsageException e) {
       err.print(e.getMessage());
-      return 1;
+      // We return exit code 2 to differentiate usage issues from code formatting issues.
+      return 2;
     } finally {
       err.flush();
       out.flush();
--- a/core/src/main/java/com/google/googlejavaformat/java/Main.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Main.java
@@ -79,7 +79,7 @@ public final class Main {
       return formatter.format(args);
     } catch (UsageException e) {
       err.print(e.getMessage());
-      return 0;
+      return 1;
     } finally {
       err.flush();
       out.flush();


core/src/main/java/com/google/googlejavaformat/java/Main.java
```diff
@@ -79,7 +79,7 @@ static int main(PrintWriter out, PrintWriter err, String... args) {
       return formatter.format(args);
     } catch (UsageException e) {
       err.print(e.getMessage());
-      return 0;
+      return 1;
```
tbroyer, line 82: Built-in commands use a status of 2 for usage issues. IIUC, this could use 2 to align (although any non-zero is enough)

abstractlyZach: I wasn't aware of certain codes being used for specific scenarios; I just thought each command assigned their own meanings to non-zero codes.

After reading https://tldp.org/LDP/abs/html/exitcodes.html, `2` seems like a good fit, but it also seems like it's reserved for shell builtins?

I don't have much of an opinion on the particular number here. Happy to defer to anyone who has experience in this area

tbroyer: Well, Picocli uses 2 for usage errors as well (https://picocli.info/#_exception_exit_codes), but that's a detail and 1 works just as well (in the case of GJF though, 1 can also mean an error happened formatting a file, using another code would allow distinguishing those cases)

abstractlyZach: Cool, I've updated with exit code 2 and included a comment

cushon (approved):

PR #927, opened 2023-03-28
Update README.md

erin-flannery-glean:
Updated the way to find Custom VM Option for Intellij IDEA Community Edition 2022.3.3

Custom VM Option cannot be searched from the help menu. 

Screenshot 2023-03-28 at 12 01 41 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ actions.
 #### IntelliJ JRE Config
 
 The google-java-format plugin uses some internal classes that aren't available
-without extra configuration. To use the plugin, go to `Help→Edit Custom VM
-Options...` and paste in these lines:
+without extra configuration. To use the plugin, find the action `Help→Find Action`. Then search for "Custom VM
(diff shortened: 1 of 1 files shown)


cushon:
I see it in the help menu (it's the second from last entry), although it doesn't show up when I type the name to search for it. Does that match what you're seeing?

Screenshot 2023-03-31 at 12 05 54 PM

erin-flannery-glean:
> I see it in the help menu (it's the second from last entry), although it doesn't show up when I type the name to search for it. Does that match what you're seeing?
> 
> Screenshot 2023-03-31 at 12 05 54 PM

Yes that matches up!

plumpy:
Okay, I'm going to go ahead and close this then. I don't know why searching doesn't work, but that seems like a JetBrains issue. The instructions as written are correct and are simpler to follow.

Thanks for reporting this, though!

PR #971, opened 2023-09-15
`GoogleJavaFormatImportOptimizer`: re-run import format if `document.txt` has changed

facboy:
This is a follow up to #960 - occasionally the imports are left unformatted and can be annoying.


plumpy:
Are you sure this is what's happening? Can you reproduce this?

My plan was that if we detected this case, I would re-trigger the `ImportOptimizer`, so it would call `processFile` again (off the Swing thread) and then call the returned runnable back on the Swing thread.

But as far as I can tell IntelliJ already handle that for us. What I see when I try to reproduce this is that IntelliJ is smart and basically if you run "optimize imports" and then change the file before it has a chance to call the returned runnable and make the edit, it knows the file changed, and it will throw that Runnable away and call `processFile` and start the whole process over again.

So I'm wondering if the behavior you're seeing has a different cause?

facboy:
it's pretty sporadic, i can revert back to the released version and see.  anedoctally, i haven't seen it once since i fixed this locally.

Issue #983, opened 2023-10-30
Google Java Format fails on complex switch expressions with Java 21


bmeier-pros:
Version 1.18.1 formatter throws an exception when trying to handle code like this:
```java
      return switch (obj1) {
          case null -> obj2 == null;
          case T1 myT1 when obj2 instanceof T1 otherT1 -> myT1.equalsStrict(otherT1);
          // exception ^ at the when (69:46 in the original source)
          case T2 myT2 when obj2 instanceof T2 otherT2 -> T2.isDeepEquals(myT2, otherT2);
          default -> obj1.equals(obj2);
      };
```

The exception is:
```
Step 'google-java-format' found problem in 'src/test/java/com/example/test/acceptance/ExpectedResult.java':
69:46: error: : or -> expected
com.google.googlejavaformat.java.FormatterException: 69:46: error: : or -> expected
        at com.google.googlejavaformat.java.FormatterException.fromJavacDiagnostics(FormatterException.java:51)
        at com.google.googlejavaformat.java.Formatter.format(Formatter.java:149)
        at com.google.googlejavaformat.java.Formatter.getFormatReplacements(Formatter.java:273)
        at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:247)
        at com.google.googlejavaformat.java.Formatter.formatSource(Formatter.java:213)
        at com.diffplug.spotless.glue.java.GoogleJavaFormatFormatterFunc.apply(GoogleJavaFormatFormatterFunc.java:65)
        at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:32)
        at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:82)
        at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:103)
        at com.diffplug.spotless.Formatter.compute(Formatter.java:246)
        at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:203)
        at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:190)
        at com.diffplug.gradle.spotless.SpotlessTaskImpl.processInputFile(SpotlessTaskImpl.java:105)
        at com.diffplug.gradle.spotless.SpotlessTaskImpl.performAction(SpotlessTaskImpl.java:89)
<gradle stack trace elided>
```

cushon:
[...]

PR #988, opened 2023-11-10
Add support for guard clauses in Java 21 switch expressions

TheCK:
This PR adds support for `switch` statements where a `case` has a guard clause.

See Issue #983

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/Formatter.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/Formatter.java
@@ -151,7 +151,17 @@ public final class Formatter {
     OpsBuilder builder = new OpsBuilder(javaInput, javaOutput);
     // Output the compilation unit.
     JavaInputAstVisitor visitor;
-    if (Runtime.version().feature() >= 17) {
+    if (Runtime.version().feature() >= 21) {
+      try {
+        visitor =
+            Class.forName("com.google.googlejavaformat.java.java21.Java21InputAstVisitor")
+                .asSubclass(JavaInputAstVisitor.class)
+                .getConstructor(OpsBuilder.class, int.class)
+                .newInstance(builder, options.indentationMultiplier());
+      } catch (ReflectiveOperationException e) {
+        throw new LinkageError(e.getMessage(), e);
+      }
+    } else if (Runtime.version().feature() >= 17) {
       try {
         visitor =
             Class.forName("com.google.googlejavaformat.java.java17.Java17InputAstVisitor")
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/google/googlejavaformat/java/java21/Java21InputAstVisitor.java
```diff
@@ -0,0 +1,76 @@
...
+        if (!first) {
+          token(",");
+          builder.breakOp(" ");
+        }
+        scan(expression, null);
+        first = false;
+      }
+      builder.close();
+    }
+    if (node.getGuard() != null) {
```
cushon, line 41: Since this block is the only part that needs to be different between 17 and 21, I would like to avoid duplicating the rest of the method.

An alternative would be to do something like this: https://github.com/google/google-java-format/blob/205b85f2315c48b2d6280a18c6ea9dbe814d3709/core/src/main/java/com/google/googlejavaformat/java/JavaInputAstVisitor.java#L3781

There could be a `getGuard` method that returns null in the base class, and `Java21InputAstVisitor` could override it

core/src/main/java/com/google/googlejavaformat/java/Formatter.java
```diff
@@ -151,7 +151,17 @@ public CharSequence getCharContent(boolean ignoreEncodingErrors) throws IOExcept
...
     // Output the compilation unit.
     JavaInputAstVisitor visitor;
-    if (Runtime.version().feature() >= 17) {
+    if (Runtime.version().feature() >= 21) {
+      try {
+        visitor =
+            Class.forName("com.google.googlejavaformat.java.java21.Java21InputAstVisitor")
+                .asSubclass(JavaInputAstVisitor.class)
+                .getConstructor(OpsBuilder.class, int.class)
+                .newInstance(builder, options.indentationMultiplier());
```
cushon, line 160: It might be worth creating a shared helper to reflectively instantiate the visitor at this point:

```
private static JavaInputAstVisitor createVisitor(String className) { .. }
```

cushon (changes requested):
Thanks for the fix!

Could you add a test? There are some examples of tests that are only run on newer Java versions here: https://github.com/google/google-java-format/blob/205b85f2315c48b2d6280a18c6ea9dbe814d3709/core/src/test/java/com/google/googlejavaformat/java/FormatterIntegrationTest.java#L49

cushon (approved):

TheCK:
@cushon I've implemented your requested changes. Hope this works for you. If you think it needs more attention still, let me know.

PR #996, opened 2023-12-02
Bump Guava to 32.1.3

Goooler:
There is an issue about resolving transitive dependencies in Guava breaks GJF 1.18.1 integration in [Spotless](https://github.com/diffplug/spotless), it has been fixed in Guava 32.1.3, it would be nice to follow this update.

See:
- https://github.com/google/guava/issues/6657.
- https://github.com/google/guava/releases/tag/v32.1.3.


moritzluedtke:
When will the fix be released? Our builds are all broken when we use the newest version of spotless (`6.23.3`).

PR #1016, opened 2023-12-19
chore: bump checkout/setup-java to v4 (use nodejs20 runtime)

chenrui333:
- bump to latest v4 (which uses nodejs 20 runtime)

cc @cushon

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
         with:
           access_token: ${{ github.token }}
       - name: 'Check out repository'
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
       - name: 'Set up JDK ${{ matrix.java }} from jdk.java.net'
         if: ${{ matrix.java == 'EA' }}
         uses: oracle-actions/setup-java@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -45,13 +45,12 @@ jobs:
             experimental: true
     runs-on: ${{ matrix.os }}
     continue-on-error: ${{ matrix.experimental }}
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
```
cushon, line 50: I'm seeing the presubmit for this PR is cancelling all but one of the actions with the message

> Canceling since a higher priority waiting request for 'CI-refs/pull/1016/merge' exists

I think maybe the issue is that the actions for different JDK versions are using the same concurrency group, and then `cancel-in-progress` cancels all but one of the them?

cushon (changes requested):

.github/workflows/ci.yml
chenrui333: let me switch this to a different PR, can you also attach the run logs for me to debug?

cushon (approved):

PR #1044, opened 2024-02-09
Add Maven Wrapper

vorburger:
@cushon how about this?

See https://maven.apache.org/wrapper/

This is sorta like https://github.com/bazelbuild/bazelisk (-ish).

Also makes it more reproducible (the project, instead of the user's environment, or GitHub image, controls the Maven version, through `.mvn/wrapper/maven-wrapper.properties`).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
-mvnw.cmd binary
+mvnw     text eol=lf
+mvnw.cmd text eol=crlf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+mvnw.cmd binary
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,10 +68,10 @@ jobs:
           cache: 'maven'
       - name: 'Install'
         shell: bash
-        run: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+        run: ./mvnw install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
       - name: 'Test'
         shell: bash
-        run: mvn test -B
+        run: ./mvnw test -B
 
   publish_snapshot:
     name: 'Publish snapshot'
@@ -94,4 +94,4 @@ jobs:
         env:
           CI_DEPLOY_USERNAME: ${{ secrets.CI_DEPLOY_USERNAME }}
           CI_DEPLOY_PASSWORD: ${{ secrets.CI_DEPLOY_PASSWORD }}
-        run: mvn -pl '!eclipse_plugin' source:jar deploy -B -DskipTests=true -Dinvoker.skip=true -Dmaven.javadoc.skip=true
+        run: ./mvn -pl '!eclipse_plugin' source:jar deploy -B -DskipTests=true -Dinvoker.skip=true -Dmaven.javadoc.skip=true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,9 +30,9 @@ jobs:
 
       - name: Bump Version Number
         run: |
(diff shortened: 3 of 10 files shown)


cushon:
The `gradlew` / `bazelisk` model seems like a good approach, but on the other hand I don't remember running into differences between mvn versions that mattered for this build. Is this required for #1045?

vorburger:
> but on the other hand I don't remember running into differences between mvn versions that mattered for this build.

More so than "problems", this means you don't have to have Maven pre-installed to build it, which seems handy?

> Is this required for #1045?

Yeah, I'm using it there. In theory, without this, the `Dockerfile` from that PR could also just install Maven itself somehow, of course; but this approach seemed better, to me.

vorburger:
@cushon what's your verdict on this one?

I still think this "lowers the entry barrier", and can't see much downside/drawback?

vorburger:
@cushon do you now (after the other PRs were merged) want me to rebase (and make required adaptions), or just abandon this one, if you are not a big fan of merging this? (Absolutely no offense taken!) Are there any other active maintainers of this project who may want to chime in? @plumpy @cpovirk @nreid260 perhaps?

cushon:
Let me mull it over. Lowering the barrier to entry is a good thing , but it's not clear to me how much this does that, and it feels a bit like clutter. I am open to being convinced this is become a best practice for maven repos and I should get with the times. I see that Guava has [recently adopted mvnw](https://github.com/google/guava/commit/5550ffc0e04fe8a3e8f128f692d37397232dae2b), maybe I'll ask them how that's going after the weekend.

vorburger:
https://github.com/search?q=mvnw.cmd&type=code may be a "data point" of
interest in this context?

cushon:
> https://github.com/search?q=mvnw.cmd&type=code may be a "data point" of interest in this context?

Thanks, it looks like there are about [1.5k `mvnw` files](https://github.com/search?q=path%3A%2F%5Emvnw%24%2F&type=code) on github, and [>500k `pom.xml` files](https://github.com/search?q=path%3A%2F%5Epom.xml%24%2F&type=code)

cpovirk:
mvnw/gradlew/bazelisk certainly "feels like" the right thing to me, but I have little concrete evidence about mvnw in particular. (I have an impression, _very possibly incorrect_, that Maven is relatively stable in comparison to the other two, at least in the projects I'm familiar with.)

I can only say that:
- Our motivation was [a report of a build error with a specific version of Maven](https://groups.google.com/g/guava-discuss/c/e5UVvuM9CP8/m/w2H3CjZ7AQAJ).
- I don't recall any reports of problems from our adding `mvnw`.
- I do worry than a Maven version in config file will go out of date more quickly than the versions on our workstations and in GitHub Actions, since Dependabot doesn't currently update the config file: https://github.com/dependabot/dependabot-core/issues/485

cpovirk:
It did occur to me that I think we've sometimes had to upgrade "things" in order to build with newer versions of Java. However, I'm not sure if we've ever needed to update more than _plugins_. Granted, upgrading Maven will give you (IIUC) some plugin updates for free. So if we're going to pin our Maven version, we might want to make sure that we're setting explicit plugin versions in our `pom.xml` files (which is possibly a good idea, anyway, so that the version of Maven affects one less thing) and having Dependabot or some other tool update them periodically.

Or we just wait for things to break and then upgrade at that point :)

(Honestly, I'm kind of fuzzy on the exact relationship between Maven and its plugins. Like, besides the fact that upgrading Maven can give you new default versions of plugins, but can it also affect which versions of plugins can even _work_ with your Maven version? I thought I remembered having to tweak versions of _dependencies_ of plugins before, similar to [this old `plexus-io` change](https://github.com/google/truth/commit/c8839658ba1fccb30c74f8cae90f21fd991a6598), perhaps related to Maven versions. I don't think I actually care much about the answer here; I'm just recording my ignorance.)

PR #1048, opened 2024-02-10
Build native binary in CI (fixes google#894)

vorburger:
Alternative for #1045

Fixes #894

PR #1049, opened 2024-02-10
Mechanically reformat GitHub Actions YAML

vorburger:
https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions proposes this format as the "canonical" one.

Adopting this makes it easier to contribute to this project by using that VSC Extension as an editor.

This change is purely mechanical and contains no other "semantic" differences.

PR #1050, opened 2024-02-10
Add missing ASL (license) to GitHub Actions

vorburger:
@cushon merge this? Just a minor clean-up, to make #1048 shorter and easier to review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-# Copyright 2020 The Error Prone Authors.
+# Copyright 2020 The Google Java Format Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


vorburger:
This will look "cleaner" to review after a rebase if you are willing to merge #1049, first.

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Error Prone Authors.
+# Copyright 2020-2024 The Google Java Format Authors
```
cushon, line 1: This is a nit, but I think the guidance is to use the year the file was created, and that it doesn't have to be updated to the current year, so just '2020' should be sufficient.

cushon (changes requested):
Thanks, looks good aside from maybe adjusting the years (in both files).

.github/workflows/ci.yml
vorburger: sure thing, done!

PR #1053, opened 2024-02-11
Build MacOS native binary on CI

vorburger:
Follow-up to #1048 to see if if so happens to "just work" on MacOS... it just might? Let's try and see!

PS: I'll tackle Windows separately, because that won't (because of the bash shell test script).

@cushon


vorburger:
FTR: This was done as part of #1055, a little differently.

I'll therefore close this PR.

PR #1054, opened 2024-02-11
Optimze CI to not re-build Eclipse Plugin etc. on GraalVM Native Runner

vorburger:
@cushon would this make sense?

Re-building the Eclipse plugins and re-running the tests on GraalVM instead of OpenJDK isn't all that interesting - is it?

Just to make it build slightly faster on CI - and save some 10^-♾️ MW energy for the 🪐  planet!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,9 +78,11 @@ jobs:
           native-image-job-reports: "true"
           cache: "maven"
       - name: "Install"
+        if: ${{ matrix.java != 'GraalVM' }}
         shell: bash
         run: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
       - name: "Test"
+        if: ${{ matrix.java != 'GraalVM' }}
         shell: bash
         run: mvn test -B
       - name: "Native"


.github/workflows/ci.yml
```diff
@@ -78,9 +78,11 @@ jobs:
           native-image-job-reports: "true"
           cache: "maven"
       - name: "Install"
+        if: ${{ matrix.java != 'GraalVM' }}
         shell: bash
         run: mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
       - name: "Test"
+        if: ${{ matrix.java != 'GraalVM' }}
```
cushon, line 85: Skipping these steps for GraalVM sounds good.

I'm realizing that there isn't a lot of overlap between the steps that we want to do for GraalVM and other JDKs--I think all of the steps here happen on other JDKs, or on GraalVM, but not both.

What do you think about adding a separate entry under `jobs:` for the GraalVM test, isntead of adding it to the matrix with the other JDKs and avoiding all of the checks for `java == graal` and `java != graal`?

vorburger: FTR: This was done in #1055. (I'll close this PR.)

vorburger:
FTR: This was done in #1055, a little differently.

I'll therefore close this PR.

PR #1055, opened 2024-02-11
Refactor CI to run GraalVM native build in separate job

vorburger:
As [proposed here](https://github.com/google/google-java-format/pull/1054#discussion_r1485659262) by @cushon

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,14 +81,14 @@ jobs:
         shell: bash
         run: mvn test -B
 
-  test-OpenJDK:
+  test-GraalVM:
     name: "GraalVM on ${{ matrix.os }}"
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.experimental }}
+    continue-on-error: true
     steps:
       - name: Cancel previous
         uses: styfle/cancel-workflow-action@0.9.1
@@ -111,7 +111,7 @@ jobs:
 
   publish_snapshot:
     name: "Publish snapshot"
-    needs: test
+    needs: test-OpenJDK
     if: github.event_name == 'push' && github.repository == 'google/google-java-format' && github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,13 +23,13 @@ on:
       - master
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -83,6 +80,31 @@ jobs:
...
         run: mvn test -B
+
+  test-OpenJDK:
+    name: "GraalVM on ${{ matrix.os }}"
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
```
cushon, line 91: `experimental` isn't part of the matrix for this job, I think it needs to be added, or else this could just hard-code `continue-on-error: true` for now if we want it to be true for both jobs

We should probably also remove `continue-on-error: true` once we start doing releases that depend on this, but I think it's OK to leave in for now while the graal stuff is being iterated on

cushon:
Thanks! This looks about like what I expected, but I don't see see any CI runs for this PR? https://github.com/google/google-java-format/actions/workflows/ci.yml

I'm not sure why it hasn't run, any ideas? I wonder if there's an error here preventing it from running that's not being reported.

cushon:
Ah, neat, the built-in GitHub editor shows errors for the config file. I pushed a commit that I think fixes it.

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 0c8cc2b5..322f914e 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,14 +81,14 @@ jobs:
         shell: bash
         run: mvn test -B
 
-  test-OpenJDK:
+  test-GraalVM:
     name: "GraalVM on ${{ matrix.os }}"
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
-    continue-on-error: ${{ matrix.experimental }}
+    continue-on-error: true
     steps:
       - name: Cancel previous
         uses: styfle/cancel-workflow-action@0.9.1
@@ -111,7 +111,7 @@ jobs:
 
   publish_snapshot:
     name: "Publish snapshot"
-    needs: test
+    needs: test-OpenJDK
     if: github.event_name == 'push' && github.repository == 'google/google-java-format' && github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     steps:
```

cushon:
CI looks better now, I'll go ahead and merge but we can keep iterating if you disagree with any of the edits

Issue #868, opened 2022-12-09
Release native builds


dlicois:
Hello

GraalVM native-image support appears to have been added a little while ago https://github.com/google/google-java-format/pull/728 (see also https://github.com/google/google-java-format/issues/358). However there are no official native build release on the release page https://github.com/google/google-java-format/releases. is it working ?

Providing these could greatly benefit to java tooling as it would allow fast formatting without having to rely on IDEs customization and while being fully compliant with google style guide. This could be easily imported in Visual Studio Code for example.

Platform targets of most interest would be win32-x64, linux-x64, darwin-x64, darwin-arm64.

vorburger:
Yeah, this would be nice. 

Especially as there are now people who distribute native builds themselves elsewhere.

Before thinking "release", there should probably be a way to (easily) build it here in this project? 

So #894 would probably be a pre-requisite here.

vorburger:
My #1045 will (should, to be tested and confirmed) make a native builds available at the next release cycle!

vorburger:
As (sorry for cross posting, I'm just excited).

FYI https://github.com/JoseVSeb/google-java-format-for-vs-code/issues/17 with links to my posts about this:

* https://tooting.ch/@vorburger/111952897874900517
* https://bsky.app/profile/vorburger.ch/post/3klp4ip2yqc2k
* https://twitter.com/vorburger/status/1759220021568340105

_"Formatting completed in 45ms."_ instead of 473ms previously...

That x10 is actually noticeable on Ctrl-S Save a *.java, even on a fast computer - very cool!

@cushon close this issue?

natros:
You can get better results if you compile with profile-guided optimizations (PGO)

vorburger:
> You can get better results if you compile with profile-guided optimizations (PGO)

@natros we welcome your PR to the configuration that's [here](https://github.com/google/google-java-format/blob/92c609a43bc894e4cd69e9c037491163303d97ed/core/pom.xml#L308) if you have expertise in this and would like to contribute that!

cushon:
I'm going to close this out for now, PGO sounds worth considering as a follow-up.

PR #1057, opened 2024-02-13
Generate native-image binaries for google-java-format

cushon:
And include them in release artifacts.

I tested this on my fork of the repo, and have a demo here: https://github.com/cushon/google-java-format/releases

When downloading the artifacts from the releases page they don't have the executable bit set, and my Mac blocks them because they aren't signed. That can be worked around with the following, but isn't ideal:

```
chmod a+rx google-java-format-darwin
sudo xattr -r -d com.apple.quarantine google-java-format-darwin
```

Progress towards #868, builds on @vorburger's work in #1048.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,7 +88,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     env:
-      SUFFIX: ${{fromJson('{"ubuntu-latest":"linux-x86_64", "macos-latest":"darwin-arm64", "windows-latest":"windows_x86-64"}')[matrix.os]}}
+      SUFFIX: ${{fromJson('{"ubuntu-latest":"linux-x86-64", "macos-latest":"darwin-arm64", "windows-latest":"windows-x86-64"}')[matrix.os]}}
     steps:
       - name: "Check out repository"
         uses: actions/checkout@v4
@@ -103,9 +103,9 @@ jobs:
       - name: "Native"
         run: mvn -Pnative -DskipTests package -pl core -am
       - name: "Move outputs"
-        run: cp core/target/google-java-format google-java-format-${{ env.SUFFIX }}
+        run: cp core/target/google-java-format google-java-format_${{ env.SUFFIX }}
       - name: "Upload native-image"
         env:
           GH_TOKEN: ${{ github.token }}
           GH_REPO: ${{ github.repository }}
-        run: gh release upload "v${{ github.event.inputs.version }}" "google-java-format-${{ env.SUFFIX }}"
+        run: gh release upload "v${{ github.event.inputs.version }}" "google-java-format_${{ env.SUFFIX }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,6 +87,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+    env:
+      SUFFIX: ${{fromJson('{"ubuntu-latest":"linux-x86_64", "macos-latest":"darwin-arm64", "windows-latest":"windows_x86-64"}')[matrix.os]}}
     steps:
       - name: "Check out repository"
         uses: actions/checkout@v4
@@ -101,9 +103,9 @@ jobs:
       - name: "Native"
         run: mvn -Pnative -DskipTests package -pl core -am
       - name: "Move outputs"
-        run: cp core/target/google-java-format google-java-format-${{ matrix.os == 'ubuntu-latest' && 'linux' || 'darwin' }}
+        run: cp core/target/google-java-format google-java-format-${{ env.SUFFIX }}
       - name: "Upload native-image"
         env:
           GH_TOKEN: ${{ github.token }}
           GH_REPO: ${{ github.repository }}
-        run: gh release upload "v${{ github.event.inputs.version }}" "google-java-format-${{ matrix.os == 'ubuntu-latest' && 'linux' || 'darwin' }}"
+        run: gh release upload "v${{ github.event.inputs.version }}" "google-java-format-${{ env.SUFFIX }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,7 @@ jobs:
     needs: build-maven-jars
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


cushon:
@vorburger can I get your review of this?

.github/workflows/release.yml
```diff
@@ -79,3 +79,31 @@ jobs:
...
             core/target/google-java-format*
             eclipse_plugin/target/google-java-format-eclipse-plugin-*.jar
+
+  build-native-image:
+    name: "Build GraalVM native-image on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    needs: build-maven-jars
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
```
vorburger, line 89:
```suggestion
        os: [ubuntu-latest, macos-latest, windows-latest]
```

.github/workflows/release.yml
```diff
@@ -79,3 +79,31 @@ jobs:
...
+        with:
+          java-version: "21"
+          distribution: "graalvm-community"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          native-image-job-reports: "true"
+          cache: "maven"
+      - name: "Native"
+        run: mvn -Pnative -DskipTests package -pl core -am
+      - name: "Move outputs"
+        run: cp core/target/google-java-format google-java-format-${{ matrix.os == 'ubuntu-latest' && 'linux' || 'darwin' }}
```
vorburger, line 104: @cushon I don't understand this expression... and it (now) won't work for Windows?

vorburger (changes requested):

.github/workflows/release.yml
cushon: It returns the string 'linux' or 'darwin' depending on the current OS, so the platform-specific outputs get uploaded with unique names. The syntax is documented here: https://docs.github.com/en/actions/learn-github-actions/expressions

I agree we'll want to update this for windows, I'm not sure what the cleanest way to do that is yet. I looked around at a few different ways to upload these artifacts to the release and didn't find anything very appealing.

vorburger: Oh, I think I just got it - the idea is that this will evaluate to the following, right?

* `google-java-format-linux`
* `google-java-format-darwin`

How about we use slightly different suffixes for the released native binaries? Inspired by e.g. `uname -a` output, I would suggest:

* `google-java-format_linux-x86_64`
* `google-java-format_darwin-arm64` (or something like that - what exactly does `uname -a` output on your Mac?)
* `google-java-format_win-x64` (I think that's what that's commonly referred to - but I'm not 100% sure)

cushon: Those suffixes LGTM, thanks. (My uname output includes `Darwin` and `arm64`.)

vorburger: > I agree we'll want to update this for windows, I'm not sure what the cleanest way to do that is yet. I looked around at a few different ways to upload these artifacts to the release and didn't find anything very appealing.

Simplest could be to duplicate this (and the "gh release upload") lines x3, with an "if" for penguins, apples, and wins?

vorburger (approved):
LGTM from me, if you wanted to already merge this as-is?

.github/workflows/release.yml
vorburger: > * `google-java-format_win-x64` (I think that's what that's commonly referred to - but I'm not 100% sure; but I would _NOT_ use "x86_64" à la Linux for this binary, because MS seems to use x86=32bit VS x64=64bit - apparently.)

Just noticed on https://github.com/bazelbuild/bazel/releases that they use "*windows_x86-64" so perhaps we can go for that here as well after all.

cushon: Thanks! I found a different way to express that mapping, and updated to use the suggested names.

.github/workflows/release.yml
```diff
@@ -87,6 +87,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+    env:
+      SUFFIX: ${{fromJson('{"ubuntu-latest":"linux-x86_64", "macos-latest":"darwin-arm64", "windows-latest":"windows_x86-64"}')[matrix.os]}}
```
vorburger, line 91: Fancy!! That's a really neat trick - TIL; Thank You!

.github/workflows/release.yml
```diff
@@ -79,3 +79,33 @@ jobs:
...
+
+  build-native-image:
+    name: "Build GraalVM native-image on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    needs: build-maven-jars
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    env:
+      SUFFIX: ${{fromJson('{"ubuntu-latest":"linux-x86_64", "macos-latest":"darwin-arm64", "windows-latest":"windows_x86-64"}')[matrix.os]}}
```
vorburger, line 91:
```suggestion
      SUFFIX: ${{fromJson('{"ubuntu-latest":"linux-x86-64", "macos-latest":"darwin-arm64", "windows-latest":"windows-x86-64"}')[matrix.os]}}
```

.github/workflows/release.yml
```diff
@@ -79,3 +79,33 @@ jobs:
...
+        with:
+          java-version: "21"
+          distribution: "graalvm-community"
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          native-image-job-reports: "true"
+          cache: "maven"
+      - name: "Native"
+        run: mvn -Pnative -DskipTests package -pl core -am
+      - name: "Move outputs"
+        run: cp core/target/google-java-format google-java-format-${{ env.SUFFIX }}
```
vorburger, line 106:
```suggestion
        run: cp core/target/google-java-format google-java-format_${{ env.SUFFIX }}
```

.github/workflows/release.yml
```diff
@@ -79,3 +79,33 @@ jobs:
...
+      - name: "Upload native-image"
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: gh release upload "v${{ github.event.inputs.version }}" "google-java-format-${{ env.SUFFIX }}"
```
vorburger, line 111:
```suggestion
        run: gh release upload "v${{ github.event.inputs.version }}" "google-java-format_${{ env.SUFFIX }}"
```

vorburger (approved):
Proposed changes, you'll also want to rebase this after merging #1060; otherwise LGTM!

PR #1059, opened 2024-02-13
Fix chmod +x util/test-native.sh which Copybara broke

vorburger:
@cushon


cushon (approved):
Thanks for the fix. I’m curious see if the import handles this correctly, if it doesn’t we can fix the permissions internally.

vorburger:
> Thanks for the fix. I’m curious see if the import handles this correctly, if it doesn’t we can fix the permissions internally.

What I can provide as input here is that it **_WAS_** correct in #1048 ...

Just to clarify that this wasn't a wrong file mode, it was that somehow the filemode got "lost" by Copybara.

I'll admit that I don't fully understand what #1058 did - it seems like that was "empty", so that shouldn't have this?

Could this be a bug with handling file permissions in Copybara?

PR #1060, opened 2024-02-13
Add Windows native build

vorburger:
@cushon how about this?

~~PS: This PR includes 2 commits, the 1st one is just #1059 - perhaps merge that first, and then rebase this?~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     continue-on-error: true
     steps:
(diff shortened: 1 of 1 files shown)


vorburger:
Huh, the JavaDoc generation fails, ONLY on GraalVM ONLY on Windows - this is fun!

#1061 may fix this? I'll rebase this after that's merged and we'll see...

cushon:
Looks good. Can you rebase to pick up the javadoc plugin upgrade, to confirm that fixes it?

vorburger:
> Looks good. Can you rebase to pick up the javadoc plugin upgrade, to confirm that fixes it?

Done! And it's green. Merge?

vorburger:
@cushon merge this?

.github/workflows/ci.yml
```diff
@@ -104,8 +104,11 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           native-image-job-reports: "true"
           cache: "maven"
-      - name: "Native"
-        run: mvn -Pnative -DskipTests package -pl core -am && util/test-native.sh
+      - name: "Native Build"
+        run: mvn -Pnative -DskipTests package -pl core -am
+      - name: "Native Test"
+        if: ${{ matrix.os != 'windows-latest' }}
+        run: util/test-native.sh
```
cushon, line 111: To check my understanding, we're skipping this because the test script isn't compatible with Windows, but we expect the native-image output still will be, right?

Maybe add a comment?

vorburger: > To check my understanding, we're skipping this because the test script isn't compatible with Windows, but we expect the native-image output still will be, right?

Yup, that Bash script for testing won't work on Windows. Someone could write a `*.bat` or `*.ps1` equivalent later, but I'm not up for that - let's get this in as-is (from my side).

> Maybe add a comment?

Done!

cushon: Thanks!

cushon (approved):

PR #1139, opened 2024-08-20
Add required jvmArgs to runIde for local development testing

nrayburn-tech:
When running locally you aren't able to customize the VM options.  This adds the arguments to the `runIde` task, so that the plugin can be tested locally.


plumpy:
Thanks for this, since the `build.gradle.kts` file now uses the newer version of the IntelliJ plugin, this won't merge cleanly anymore. So I added it in the version upgrade commit: 5544952d819554d040608462a46e7185edc2b216

Issue #1087, opened 2024-03-20
Automatic deindentation of long text blocks


bmeier-pros:
I have a file that consists mostly of constant strings written as text blocks.  When I apply the formatter to the file, the text blocks are inconsistently indented, with some being completely left justified, and others being shifted underneath the opening """.  There doesn't seem to be any particular rationale for which blocks get justified, but it is consistent.  I'm trying to generate a sanitized reproduction currently.

Stephan202:
Hey @bmeier-pros! Does your issue match what's described in #1081?

bmeier-pros:
@Stephan202 - not completely, I saw that issue as well, but these are just bare `static final string` constants.  

For example:

```java
    public static final String JSON_1 =
        """
        [
          "FUNCTION1",
          "FUNCTION2",
          "// ... 98 other records ..."
          "FUNCTION101"
        ]""";

    public static final String JSON_2 =
        """
{
  "errorID": "error id",
  "errorMessage": "message",
  "resourceType": "metadata",
  "requestID": "request"
}""";
```

I'd say it was something about the JSON curly brackets, but it's not consistent.  I'm trying to find a truly minimal reproduction.

cushon:
I suspect the behaviour you're seeing is that there's some logic to start the text block contents at the left margin if indenting the lines would cause the column limit to be exceeded, and to preserve the existing formatting choice if a text block is aligned with the left margin.

In that example, adding a least one space of indent to the second text block will cause the formatter to indent it the same way as the first text block.

The formatting of text blocks hasn't been finalized yet, we might revisit some of those details, in particular whether or not aligning with the left margin is a choice that should be preserved.

https://github.com/google/google-java-format/blob/71a755b4daf0846c867834b5a0d86297ac7d0667/core/src/main/java/com/google/googlejavaformat/java/StringWrapper.java#L204-L207

Example:

```
$ google-java-format T.java
class T {
  String S =
      """
      hello
      world
      """;
}
```

```
$ google-java-format T.java
class T {
  String S =
      """
hello
world
""";
}
```

bmeier-pros:
Hi @cushon,

That makes some sense, however the original formatting of the two text blocks before applying google-java-format was consistent at 8 spaces indentation for both blocks.  This also happened in several other places throughout the file where some blocks were indented and others were justified.  It's also consistent - if I modify the indentation of the second block, it gets revised back to left justified.  I will try again to make a minimal case :-) I've had a busy couple of days.

[...]

PR #1165, opened 2024-09-24
Fix text block indentation when the text block has no prefix

dweiss:
The simplest example is this:
```
public class A {
  public void foo() {
  String a = """
lorem
ipsum
""";

  String b = """
 lorem
 ipsum
 """;
 }
}
```

Both text blocks should be aligned after the formatter pass but ```a``` isn't:
```
public class A {
  public void foo() {
    String a =
        """
lorem
ipsum
""";

    String b =
        """
        lorem
        ipsum
        """;
  }
}
```

I think the StringWrapper used to apply some more logic to actually "deindent" text block strings but that logic isn't needed and is in fact wrong - I added a comment in the patch.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/StringWrapper.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/StringWrapper.java
@@ -197,12 +197,10 @@ public final class StringWrapper {
         ImmutableList<String> initialLines = text.lines().collect(toImmutableList());
         String stripped = stripIndent(initialLines.stream().skip(1).collect(joining(separator)));
         ImmutableList<String> lines = stripped.lines().collect(toImmutableList());
-        int deindent =
-            initialLines.get(1).stripTrailing().length() - lines.get(0).stripTrailing().length();
 
         int startColumn = lineMap.getColumnNumber(startPosition);
         String prefix =
-            (deindent == 0 || lines.stream().anyMatch(x -> x.length() + startColumn > columnLimit))
+            (lines.stream().anyMatch(x -> x.length() + startColumn > columnLimit))
                 ? ""
                 : " ".repeat(startColumn - 1);
 
--- a/core/src/test/resources/com/google/googlejavaformat/java/testdata/RSLs.output
+++ b/core/src/test/resources/com/google/googlejavaformat/java/testdata/RSLs.output
@@ -48,9 +48,9 @@ ipsum
 """;
   String k =
       """
-lorem
-ipsum
-""";
+      lorem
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/google/googlejavaformat/java/StringWrapper.java
```diff
@@ -197,12 +197,10 @@ private void indentTextBlocks(
         ImmutableList<String> initialLines = text.lines().collect(toImmutableList());
         String stripped = stripIndent(initialLines.stream().skip(1).collect(joining(separator)));
         ImmutableList<String> lines = stripped.lines().collect(toImmutableList());
-        int deindent =
-            initialLines.get(1).stripTrailing().length() - lines.get(0).stripTrailing().length();
 
         int startColumn = lineMap.getColumnNumber(startPosition);
         String prefix =
-            (deindent == 0 || lines.stream().anyMatch(x -> x.length() + startColumn > columnLimit))
```
dweiss, line 205: We don't have to compute "deindent" at all here, I think, since no text block is allowed to appear at the top level of the syntax tree (?). Even if we wanted to allow it, the condition should be ```prefix = (startColumn == 0 ? "" : " ".repeat(startColumn -1)```.

core/src/test/resources/com/google/googlejavaformat/java/testdata/RSLs.output
```diff
@@ -48,9 +48,9 @@ ipsum
 """;
   String k =
       """
-lorem
-ipsum
-""";
```
dweiss, line 53: The expected output in this test was incorrect, it should have an indent.

cushon:
This is currently intentional, although there's some ongoing internal discussion about what the Google Java Style Guide should say about text blocks, and the 'outdenting' behaviour may change.

I think this is the same issue as https://github.com/google/google-java-format/issues/1087#issuecomment-2016582504

dweiss:
Yes, it seems to be the same underlying issue (code condition). I find it hard to understand why text blocks without an indent prefix would have to left without any formatting - this is inconsistent and strange to me. I also think it can lead to destructive formatting changes when used with intellij's plugin (which doesn't apply text block formatting). A subsequent run of the preformatter leaves the code in a different state compared to one run through the formatter. I'd have to come up with an example of this but I'm sure it's possible because I seem to indent those text blocks manually all the time...

cushon:
I am going to close this out for now. The 'automatic deindent' behaviour is being revisited, and I will update #1087 when that happens.

PR #1228, opened 2025-03-22
docs: Clarify impact of java version used on README

vorburger:
It at first confused the hell out of me why clearly the exact SAME version of google-java-format worked great locally but failed with errors (due to my use of record matching pattern syntax) on a GitHub Action - until I've noticed that it still ran under Java 17 - and that this matters to google-java-format (my naive initial assumption was that only the google-java-format version itself was relevant).

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ and run it with:
 java -jar /path/to/google-java-format-${GJF_VERSION?}-all-deps.jar <options> [files...]
 ```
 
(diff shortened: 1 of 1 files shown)


vorburger:
Closing this, because #1232 merged this as https://github.com/google/google-java-format/commit/14ae0a786ed7ff88e34b7b7f26c55123df997589.

PR #1415, opened 2026-07-13
Add --google-style to select Google Style

hferreiro:
--aosp was the only style switch, with no way to select Google style once it had been passed. Wrapper scripts such as Chromium's google-java-format launcher append --aosp unconditionally, leaving projects that format through them no way to opt back into Google Style.

Add a --google-style flag, and allow overriding a previously given style. When both --aosp and --google-style are given, the last one wins.

### DIFF
--- a/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
+++ b/core/src/main/java/com/google/googlejavaformat/java/CommandLineOptionsParser.java
@@ -72,6 +72,7 @@ final class CommandLineOptionsParser {
         case "--offset", "-offset" -> optionsBuilder.addOffset(parseInteger(it, flag, value));
         case "--length", "-length" -> optionsBuilder.addLength(parseInteger(it, flag, value));
         case "--aosp", "-aosp", "-a" -> optionsBuilder.aosp(true);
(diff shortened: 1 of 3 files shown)


cushon:
Thanks, this looks fine.

One minor question: WDYT about `--google-style` or something like that instead of `--no-aosp`? I don't think we're expecting to add more styles here, and the behaviour would be the same, I just wonder if it's slightly clearer.

hferreiro:
Hey @cushon, what do you think about the latest update? I followed clang-format and added a `--style=` flag instead, while keeping `--aosp` for backwards compatibility.

I'll go with `--google-style` if that's preferred.

cushon:
Thanks, I think I slightly prefer `--google-style` over `--style` since it requires less refactoring, and I'm not expecting to add more styles. (If that happens I think there'd be a strong reason to generalize it to `--style=`.)

hferreiro:
> Thanks, I think I slightly prefer `--google-style` over `--style` since it requires less refactoring, and I'm not expecting to add more styles. (If that happens I think there'd be a strong reason to generalize it to `--style=`.)

Done.

cushon (approved):

hferreiro:
@cushon can you approve the workflow run?

PR #1455, opened 2026-09-14
Fix issue 1428, Eclipse plugin commonmark dependency

mihnita:


mihnita:
Tested with the latest Eclipse (2026-09).

The current PR uses the `org.commonmark` included with Eclipse, and uses `dependency-resolution/extraRequirements` to tell the user that the `org.commonmark` plugin is required.
This seems cleaner, and we behave more like a standard Eclipse plugin.

The drawback is that we depend on the `org.commonmark` included with Eclipse,
whatever that is (0.30.0 for Eclipse 2026-09, 0.24.0 for Eclipse 2025-06, etc)

The alternative is to pack in the plugin the exact `org.commonmark` version we
tested the formatter with.
We would need the shaded jar for that, `google-java-format-{version}-all-deps.jar`.
Or "shade" the `org.commonmark` in the plugin itself.

